PR #2, opened 2017-11-02
fix getPrices for the wrapper

tranvictor:


contracts/KyberReserve.sol
```diff
@@ -335,7 +335,7 @@ contract KyberReserve {
     /// @param dest   Destinatoin token
     /// @return (conversion rate,experation block,dest token balance of reserve)
     function getPairInfo( ERC20 source, ERC20 dest ) constant returns(uint rate, uint expBlock, uint balance) {
-        ConversionRate memory rateInfo = pairConversionRate[sha3(source,dest)];
```
tranvictor: Oh removing memory is just for my experiment. You can add them back if you want.

yaronvel:
As we discussed, a better solution is if arrays reside in the memory.

PR #17, opened 2017-12-13
Ilan/add permissions contract

ilanDoron:
added mock permissions contract in mockcontracts folder.
added tests for all functions.


contracts/Permissions.sol
```diff
@@ -0,0 +1,102 @@
+pragma solidity ^0.4.0;
+
+
+contract Permissions {
+
+    address public Admin;
```
yaronvel: variable name should start with lower case letter

contracts/Permissions.sol
```diff
@@ -0,0 +1,102 @@
...
+    address public proposedAdmin;
+    mapping(address=>bool) public Operators;
```
yaronvel: also here and everywhere

contracts/Permissions.sol
```diff
@@ -0,0 +1,102 @@
+pragma solidity ^0.4.0;
+
+
+contract Permissions {
```
yaronvel: think of better name. maybe PermissionLevels? think of something better.

contracts/Permissions.sol
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    function requestOwnershipTransfer( address newAdmin ) public
+        onlyAdmin
+    {
+        proposedAdmin = newAdmin;
+    }
+
+    function acceptOwnerShip() public {
+        if (msg.sender != proposedAdmin) revert ();
```
yaronvel: change to `require`

contracts/Permissions.sol
```diff
@@ -0,0 +1,102 @@
...
+
+    modifier onlyAlerter() {
+        require (Alerters[msg.sender]);
+        _;
+    }
+
+    function requestOwnershipTransfer( address newAdmin ) public
+        onlyAdmin
+    {
+        proposedAdmin = newAdmin;
```
yaronvel: should be logged as an event

contracts/Permissions.sol
```diff
@@ -0,0 +1,102 @@
...
+        require (Operators[msg.sender]);
+        _;
+    }
+
+    modifier onlyAlerter() {
+        require (Alerters[msg.sender]);
+        _;
+    }
+
+    function requestOwnershipTransfer( address newAdmin ) public
```
yaronvel: it is not really request. think of better name. Maybe see what bancor named it.

contracts/Permissions.sol
```diff
@@ -0,0 +1,102 @@
...
+        onlyAdmin
+    {
+        proposedAdmin = newAdmin;
+    }
+
+    function acceptOwnerShip() public {
```
yaronvel: ownership is one word

contracts/Permissions.sol
```diff
@@ -0,0 +1,102 @@
...
+        if (msg.sender != proposedAdmin) revert ();
+        Admin = proposedAdmin;
```
yaronvel: log as event

contracts/Permissions.sol
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    function addAlerter( address newAlerter ) public
+        onlyAdmin
+    {
+        Alerters[newAlerter] = true;
```
yaronvel: should log as event

contracts/Permissions.sol
```diff
@@ -0,0 +1,102 @@
...
+        AlertersGroup.push(newAlerter);
```
yaronvel: should check that array size < 100. if not throw.

contracts/Permissions.sol
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    function removeAlerter ( address toRemove ) public
```
yaronvel: `toRemove` is not a good name. Maybe just call it `address alerter`.

contracts/Permissions.sol
```diff
@@ -0,0 +1,102 @@
...
+        //TODO - return success?
+        //TODO - can alerter remove himself? or only admin?
+        if (!Alerters[toRemove]) revert();
+
+        for (uint i = 0; i < AlertersGroup.length; ++i)
+        {
+            if (AlertersGroup[i] == toRemove)
+            {
+                AlertersGroup[i] = AlertersGroup[AlertersGroup.length - 1];
+                AlertersGroup.length -= 1;
```
yaronvel: there is `--` opertor

contracts/Permissions.sol
```diff
@@ -0,0 +1,102 @@
...
+        Alerters[newAlerter] = true;
+        AlertersGroup.push(newAlerter);
+    }
+
+    function removeAlerter ( address toRemove ) public
+        onlyAdmin
+    {
+        //TODO - return success?
+        //TODO - can alerter remove himself? or only admin?
+        if (!Alerters[toRemove]) revert();
```
yaronvel: log as event

contracts/Permissions.sol
```diff
@@ -0,0 +1,102 @@
...
+            if (AlertersGroup[i] == toRemove)
+            {
+                AlertersGroup[i] = AlertersGroup[AlertersGroup.length - 1];
+                AlertersGroup.length -= 1;
+                break;
+            }
+        }
+    }
+
+    function addOperator( address newOperator ) public
```
yaronvel: same comments as in alerter

contracts/Permissions.sol
```diff
@@ -0,0 +1,102 @@
...
+            if (OperatorsGroup[i] == toRemove)
+            {
+                OperatorsGroup[i] = OperatorsGroup[OperatorsGroup.length - 1];
+                OperatorsGroup.length -= 1;
+                break;
+            }
+        }
+    }
+
+    function getAlerters () public view returns (address[] alerters) {
```
yaronvel: AlertersGroup is public variable. But I am not sure what it returns. Whether it is the entire array, or just a function from index to value. Please find out. If it returns the entire array, no need for these functions.

contracts/Permissions.sol
ilanDoron: don't understand this remark

yaronvel: Do 'x--' and not 'x-=1'

PR #21, opened 2017-12-22
volume imbalance recorder

yaronvel:


yaronvel:
@ilanDoron please review

contracts/VolumeImbalanceRecorder.sol
```diff
@@ -0,0 +1,152 @@
...
+                                minimalRecordResolution,
+                                maxPerBlockImbalance,
+                                maxTotalImbalance);
+    SetTokenControlInfo( token,
+                         minimalRecordResolution,
+                         maxPerBlockImbalance,
+                         maxTotalImbalance );
+  }
+
+  function getTokenControlInfo( ERC20 token ) public view returns(uint,uint,uint) {
```
ilanDoron: in other functions we state returned parameter names. here not... what is our convention

ilanDoron:
@yaronvel  
do u have a paper describing this algorithm?

contracts/VolumeImbalanceRecorder.sol
```diff
@@ -0,0 +1,152 @@
...
+    TokenImbalanceData memory currentBlockData = tokenImbalanceData[token][currentBlockIndex];
+    // first scenario - this is not the first tx in the current block
+    if( currentBlockData.lastBlock == currentBlock ) {
+      if( uint(currentBlockData.lastPriceUpdateBlock) == priceUpdateBlock ){
+        // just increase imbalance
+        currentBlockData.lastBlockBuyUnitsImbalance += recordedBuyAmount;
+        currentBlockData.totalBuyUnitsImbalance += recordedBuyAmount;
+      }
+      else {
+        // imbalance was changed in the middle of the block
```
ilanDoron: I think remark should be - price was changed...

contracts/VolumeImbalanceRecorder.sol
```diff
@@ -0,0 +1,152 @@
...
+                                                                          currentBlock );
+    totalImbalance *= resolution;
+    currentBlockImbalance *= resolution;
+  }
+
+
+  function addImbalance( ERC20 token,
+                         int buyAmount,
+                         uint priceUpdateBlock,
+                         uint currentBlock ) internal {
```
ilanDoron: how is max maxPerBlockImbalance used. and also maxTotalImbalance.
maybe if values were crossed this function should return it as return value.

contracts/VolumeImbalanceRecorder.sol
yaronvel: convention is to be consistent, and you are right. I did mention it here so I will not forget. should remove it later.

contracts/VolumeImbalanceRecorder.sol
```diff
@@ -0,0 +1,152 @@
+pragma solidity ^0.4.18;
+
+import "./ERC20Interface.sol";
+import "./PermissionGroups.sol";
+
+interface VolumeImbalanceRecorderInterface {
```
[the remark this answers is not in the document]
yaronvel: let's discuss it. tbh, maybe 3 should be the answer.

contracts/VolumeImbalanceRecorder.sol
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  mapping(address => TokenControlInfo) tokenControlInfo;
+
+  struct TokenImbalanceData {
+    int64  lastBlockBuyUnitsImbalance;
+    uint64 lastBlock;
+
+    int64  totalBuyUnitsImbalance;
+    uint64 lastPriceUpdateBlock;
```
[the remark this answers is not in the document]
yaronvel: So everything would fit in 32 bytes. This way there is only one memory storage operation.

contracts/VolumeImbalanceRecorder.sol
```diff
@@ -0,0 +1,152 @@
...
+
+    for(uint windowInd = 0 ; windowInd < SLIDING_WINDOW_SIZE ; windowInd++ ) {
+      TokenImbalanceData memory perBlockData = tokenImbalanceData[token][windowInd];
+      if( perBlockData.lastBlock <= endBlock && perBlockData.lastBlock >= startBlock ) {
+        buyImbalance += int(perBlockData.lastBlockBuyUnitsImbalance);
+      }
+    }
+  }
+
+  function getImbalanceSincePriceUpdate( ERC20 token, uint priceUpdateBlock, uint currentBlock )
```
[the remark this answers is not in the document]
yaronvel: will be easier to test. 
The reserve manager will call it with `block.number`.

contracts/VolumeImbalanceRecorder.sol
yaronvel: in a different module, we block the tx if it exceeds max. We will verify these are not crossed before calling this function. If they are crossed we return price 0.

contracts/VolumeImbalanceRecorder.sol
yaronvel: actually both were changed in the middle. you are right.

yaronvel:
@ilanDoron regarding the algorithm.
1. imbalance = buy - sell volume.
2. we measure imbalance per block and since price was changed.
3. when price is set a block number parameter is associated to it. For example 1000.
4. 1000 is the block number when the update price tx was sent. For all we know it could have been updated only at 1004 (for example).
5. Now user made a tx only at block 1006. We update per block data, but also need to update imbalance since block 1000. For this purpose we go over all the sliding window and check.
6. It will not work properly if the set price tx took too long to be mined. But typically it should work.

ilanDoron:
consider documenting it in the contract.
did u write tests? or are there any planned tests for this?

yaronvel:
this is still work in progress. it will require a lot of testing. nothing was done already.
I am working also on the pricing module now which is very connected to it.
Pricing and this are the two most complicated modules. So I will take the lead there. Will need your help with the testing later.

contracts/VolumeImbalanceRecorder.sol
```diff
@@ -0,0 +1,152 @@
...
+    // check the imbalance in the sliding window
+    require(startBlock <= endBlock);
+
+    buyImbalance = 0;
+
+    for(uint windowInd = 0 ; windowInd < SLIDING_WINDOW_SIZE ; windowInd++ ) {
+      TokenImbalanceData memory perBlockData = tokenImbalanceData[token][windowInd];
+      if( perBlockData.lastBlock <= endBlock && perBlockData.lastBlock >= startBlock ) {
+        buyImbalance += int(perBlockData.lastBlockBuyUnitsImbalance);
+      }
```
ilanDoron: whenever range is larger then sliding window size, the results could include "old blocks" data if no block was used in the middle. but if it was the "old block" data will be erased... so in a way this is inconsistent.

contracts/VolumeImbalanceRecorder.sol
```diff
@@ -0,0 +1,152 @@
...
+    }
+    else {
+      // first tx in the current block
+      int currentBlockImbalance;
+      (prevImbalance, currentBlockImbalance) = getImbalanceSincePriceUpdate( token, priceUpdateBlock, currentBlock );
+
+      currentBlockData.lastBlockBuyUnitsImbalance = recordedBuyAmount;
+      currentBlockData.lastBlock = uint64(currentBlock);
+      currentBlockData.lastPriceUpdateBlock = uint64(priceUpdateBlock);
+      currentBlockData.totalBuyUnitsImbalance = int64(prevImbalance) + recordedBuyAmount;
```
ilanDoron: assuming we had a buy each block and that block starts from 1.
totalBuyUnitsImbalance for window index 1 will accumulate balance for blocks  , 1, 11, 21 etc.
or am I missing something?
I would expect in block 11 to have it accumulates all previous 10 (== sliding window size) blocks.

contracts/VolumeImbalanceRecorder.sol
yaronvel: it is inconsistent, and only best effort. it is possible to add some verification, but I am not sure if needed. As long as it fits into the sliding window it should be ok.

contracts/VolumeImbalanceRecorder.sol
yaronvel: you are missing something, the accumulation is not limited to the size of the sliding window. It could be infinite.
See how we use it [here](https://github.com/KyberNetwork/smart-contracts/blob/yaronvel/pricing/contracts/VolumeImbalanceRecorder.sol#L77)

contracts/VolumeImbalanceRecorder.sol
```diff
@@ -0,0 +1,152 @@
...
+                         maxTotalImbalance );
+  }
+
+  function getTokenControlInfo( ERC20 token ) public view returns(uint,uint,uint) {
+    return (tokenControlInfo[token].minimalRecordResolution,
+            tokenControlInfo[token].maxPerBlockImbalance,
+            tokenControlInfo[token].maxTotalImbalance);
+  }
+
+  function getImbalanceInRange( ERC20 token, uint startBlock, uint endBlock ) public view returns(int buyImbalance ) {
```
ilanDoron: consider making this internal - so only getImbalance is used. or some other function that multiply resolution. 
for the user its not clear enough this is should be mul with resolution.

Issue #6, opened 2017-12-11
Upgradable white list contract


yaronvel:
1. Have the interface `maxTradeSize( ERC20 inputToken, address user) view returns(uint)`.
2. `KyberNetwork` contract has a pointer to white list contract. Pointer is controlled by admin.
3. Implement white list contract.

ilanDoron:
contract name:
KyberWhiteList
parameters
----------
uint sgdToEthRate //singapore dollar
mapping - address to uint. - 	userCategory

mapping - uint to uint - categoryCap (ex: 0 starndard cap, 1 - KYC cap, etc) per category we list max sgd
some could be private cap. if some specific address has higher cap.

api
---
setUserCategory(address, uint)
setCategoryCap(uint categroy, uint sgdCap)
setSgdToEthRate(uint)
getUserCapInWei(address) returns uint

PR #27, opened 2017-12-24
Ilan/white list contract

ilanDoron:
please review.
I also added usage in kyberNetwork contract.
but this requires a change in deploy script. so how to advance with this? 
should we prepare a new deploy?


ilanDoron:
resolve #6

Issue #7, opened 2017-12-11
Upgradable get expected price function (only for UI)


yaronvel:
support `getExpectedRate( ERC20 srcToken, ERC20 destToken, uint srcQty, address userAddres )`.
`KyberNetwork` keeps a pointer to it. Upgradable by admin.

ilanDoron:
get expected rate will be a seperate contract.
1. interface ExpectedRateInterface that exposes 
- getExpectedRate
- `setKyberNetwork` to set the contract address by admin.
- 'setQuantityFactor'
2. initial ExpectedRate is Withdrawble, is PermisionGroup, is ExpectedRateInterface, KyberConstants
3. define first algorithm. getExpectedRate
bestRate = findBestRate (src, dst, quantity)
slippageRate = findBestRate(src, dst, quantity * quantityFactor) - this is to take 
return (uint bestRate, uint slippageRate)

PR #28, opened 2017-12-24
Ilan/expected rate

ilanDoron:


ilanDoron:
resolve #7

contracts/ExpectedRate.sol
```diff
@@ -0,0 +1,41 @@
+pragma solidity ^0.4.18;
+
+
+import "./ERC20Interface.sol";
+import "./KyberNetwork.sol";
+import "./PermissionGroups.sol";
+
+
+interface ExpectedRateInterface {
+    function getExpectedRate ( ERC20 source, ERC20 dest, uint destQuantity ) public view
```
yaronvel: user will never know dest qty. it should be source qty.

ilanDoron:
Yes
Of course. But
I did like this because if i imagine our future model
Let's say the user needs x omg tokens to pay for some service but it has
only ethers.
It will press pay and will know dst amount.
i realize now wallet will figure out source amount by using get expected
price.
but then. if some change is left it should be transferred to user account
or it should not be converted.
shouldnt we handle this scenario?

בתאריך 25 בדצמ 2017 17:42,‏ "yaronvel" <notifications@github.com> כתב:

ilanDoron:
i think we need the option to get expected rate both witg src quantity and
dst quantity...

בתאריך 25 בדצמ 2017 18:42,‏ "Ilan Doron" <ilan.on.the.road@gmail.com> כתב:

contracts/ExpectedRate.sol
ilanDoron: done
@yaronvel  please merge

yaronvel:
I lost track of this branch it has so many commits that seems to be unrelated.
Please pull a new branch from develop and make these changes.

ilanDoron:
yes.
my mistake
used pull instead of git rebase...

PR #62, opened 2018-01-10
Ilan/more testing

ilanDoron:
here is the sanity prices testing


contracts/SanityRates.sol
```diff
@@ -40,7 +40,7 @@ contract SanityRates is SanityRatesInterface, Withdrawable, Utils {
         uint rate;
         address token;
         if (src == ETH_TOKEN_ADDRESS) {
-            rate = (PRECISION*PRECISION)/tokenRate[dest];
+            rate = (PRECISION * PRECISION) / tokenRate[dest];
```
yaronvel: didn't we say no code changes?

PR #66, opened 2018-01-13
Ilan/more testing

ilanDoron:
notice some code changes.

### DIFF
--- a/contracts/KyberNetwork.sol
+++ b/contracts/KyberNetwork.sol
@@ -106,6 +106,7 @@ contract KyberNetwork is Withdrawable, Utils {
     function addReserve(KyberReserve reserve, bool add) public onlyAdmin {
 
         if (add) {
+            require(!isReserve[reserve]);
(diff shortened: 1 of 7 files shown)


contracts/KyberNetwork.sol
```diff
@@ -115,7 +116,8 @@ contract KyberNetwork is Withdrawable, Utils {
             for (uint i = 0; i < reserves.length; i++) {
                 if (reserves[i] == reserve) {
                     if (reserves.length == 0) return;
-                    reserves[i] = reserves[--reserves.length];
+                    reserves[i] = reserves[reserves.length - 1];
+                    --reserves.length;
```
yaronvel: why break it into two lines? also, why x++ vs ++x. let's be consistent (I think x++ is nicer).
Only reason to use --x, is if we do it in one line.

ilanDoron:
in one line it returns an opcode error...
not sure why. maybe the compiler doesn't handle it well

PR #69, opened 2018-01-15
Ilan/auditor remarks

ilanDoron:
all auditor remarks and a bit more...


contracts/KyberNetwork.sol
```diff
@@ -238,21 +248,21 @@ contract KyberNetwork is Withdrawable, Utils {
...
     /* solhint-enable code-complexity */
 
-    function getExpectedRate(ERC20 source, ERC20 dest, uint srcQuantity)
+    function getExpectedRate(ERC20 source, ERC20 dest, uint sourceQty)
         public view
         returns (uint expectedRate, uint slippageRate)
     {
-        require(expectedRateContract != address(0));
-        return expectedRateContract.getExpectedRate(source, dest, srcQuantity);
+        require(enabled);
```
yaronvel: it shouldn't check for enable. it should check for address != 0

contracts/KyberNetwork.sol
```diff
@@ -151,20 +153,28 @@ contract KyberNetwork is Withdrawable, Utils {
...
         expectedRateContract = _expectedRate;
         feeBurnerContract = _feeBurner;
         maxGasPrice = _maxGasPrice;
-        negligibleRateDiff = _negligibleDiff;
+        negligibleRateDiff = _negligibleRateDiff;
     }
 
     function setEnable(bool _enable) public onlyAdmin {
-        enable = _enable;
+        if (_enable == true) {
```
yaronvel: if(_enable)

contracts/KyberNetwork.sol
ilanDoron: but all addresses were checked upon init.
if we had a few addresses here it would be enough to check enable.

this was his remark. put all address validation in one place. preffered is constructor. but here we are limited so here its the setParams

yaronvel: I want to call it even before enable = true

ilanDoron: OK

ilanDoron:
currently we advance in premainnet branch. all code is there...

PR #86, opened 2018-01-27
adding interface to all upgradable contracts

yaronvel:
adding interface to all upgradable contracts.
This will allow upgrading them without a need to change the binary of an unchanged contract.
Hence, it will be possible to use same deployed binaries when testing upgrades.
For example, if we change `KyberReserve.sol`, the binary of `KyberNetwork.sol` would have been change before.
As a result the deployed `KyberNetwork.sol` would have different binary than the one we use for testing.

### DIFF
--- a/contracts/KyberReserve.sol
+++ b/contracts/KyberReserve.sol
@@ -4,21 +4,22 @@ pragma solidity 0.4.18;
 import "./ERC20Interface.sol";
 import "./Utils.sol";
 import "./Withdrawable.sol";
-import "./ConversionRates.sol";
+import "./ConversionRatesInterface.sol";
 import "./VolumeImbalanceRecorder.sol";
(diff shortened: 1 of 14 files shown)


contracts/KyberReserve.sol
```diff
@@ -4,21 +4,22 @@ pragma solidity 0.4.18;
 import "./ERC20Interface.sol";
 import "./Utils.sol";
 import "./Withdrawable.sol";
-import "./ConversionRates.sol";
+import "./ConversionRatesInterface.sol";
 import "./VolumeImbalanceRecorder.sol";
```
ilanDoron: while you're at this.
remove this one as well. (volume imbalance recorder...)

yaronvel: Please submit pr for that.

PR #87, opened 2018-01-27
Yaronvel/decimals

yaronvel:


contracts/Utils.sol
```diff
@@ -9,9 +9,27 @@ contract Utils {
...
+    }
+
+    function getDecimals(ERC20 token) internal view returns(uint) {
+        if (token == ETH_TOKEN_ADDRESS) return ETH_DECIMALS; // save storage access
+        uint tokenDecimals = decimals[token];
+        // technically, there might be token with decimals 0
+        // moreover, very possible that old tokens have decimals 0
+        // these tokens will just have higher gas fees.
+        if(tokenDecimals == 0) return token.decimals();
+
```
ilanDoron: I suggest if (tokenDecimals == 0) 
set decimals[token]
this way can avoid setDecimals all together. don't have to define it or call it.
a bit cleaner code imo.

yaronvel: Thought of that, but then certain functions are no longer constant.

ilanDoron: I see your point.
when ever calling a non constant function it means it can't be a local read?

yaronvel: well, the sad thing is that technically it can, because solidity compiler does not verify it. but you get compilation warnings.
So it is not a good practice.

PR #94, opened 2018-01-30
Ilan/upgrace erc20 interface

ilanDoron:
added name and symbol to ERC20Interface public function.
will enable me to verify we have the correct token.


yaronvel:
It is not part of erc20. Not sure even if we can rely on old tokens having it. Plus we dont change code now.

You can add it to mock if you want.

ilanDoron:
OK
but you also added decimals. which isn't part of ERC20

but no problem will add to mock.

PR #101, opened 2018-02-02
add validation to prevent more overflows

yaronvel:


contracts/ExpectedRate.sol
```diff
@@ -30,6 +30,8 @@ contract ExpectedRate is Withdrawable, ExpectedRateInterface {
     event MinSlippageFactorSet (uint newMin, uint oldMin, address sender);
 
     function setMinSlippageFactor(uint bps) public onlyOperator {
+        require(minSlippageFactorInBps <= 100 * 100);
+
```
ilanDoron: this looks like same maxBps const.

contracts/ConversionRates.sol
```diff
@@ -350,6 +352,10 @@ contract ConversionRates is ConversionRatesInterface, VolumeImbalanceRecorder, U
     }
 
     function addBps(uint rate, int bps) internal pure returns(uint) {
+        require(rate <= MAX_RATE);
+        require(bps >= MIN_BPS_ADJUSTMENT);
+        require(bps <= MAX_BPS_ADJUSTMENT);
+
         uint maxBps = 100 * 100;
         return (rate * uint(int(maxBps) + bps)) / maxBps;
```
ilanDoron: IMO maxBps should be a constant as well.

contracts/KyberNetwork.sol
```diff
@@ -165,6 +165,8 @@ contract KyberNetwork is Withdrawable, Utils {
         require(_whiteList != address(0));
         require(_feeBurner != address(0));
         require(_expectedRate != address(0));
+        require(_negligibleRateDiff <= 100 * 100); // at most 100%
+
```
ilanDoron: maxBps const?

yaronvel:
Ilan, please take 2nd look. Yes.

ilanDoron:
regrading u say:
Problem is max is different per application. For expected price, surely 100% diff is too much. For step function, who knows. Probably too much. But why should we limit it?

but I see 100 * 100 in 4 different places.
better git is some kind of name. maybe not maxBps

yaronvel:
Maybe. Anyway, please focus on whether more validation is required.

PR #102, opened 2018-02-03
general purpose info field in contract

yaronvel:


contracts/KyberNetwork.sol
```diff
@@ -23,7 +23,7 @@ contract KyberNetwork is Withdrawable, Utils {
     FeeBurnerInterface    public feeBurnerContract;
     uint                  public maxGasPrice = 50 * 1000 * 1000 * 1000; // 50 gwei
     bool                  public enabled = false; // network is enabled
-    uint                  public networkState; // this is only a field for UI.
+    mapping(bytes32=>uint) public info; // this is only a UI field for external app.
```
ilanDoron: don't understand why bytes32? 
what key type will be used?
how will we know per query the key?

web3deployment/deployer.js
```diff
@@ -272,7 +273,22 @@ async function main() {
...
+    var srcString1 = web3.utils.sha3("src token " + (2*i).toString());
+    var destString1 = web3.utils.sha3("dest token " + (2*i).toString());
+    var srcString2 = web3.utils.sha3("src token " + (2*i + 1).toString());
+    var destString2 = web3.utils.sha3("dest token " + (2*i + 1).toString());
+
+    await sendTx(networkContract.methods.setInfo(srcString1, ethAddress));
+    await sendTx(networkContract.methods.setInfo(destString1, tokens[i]));
+    await sendTx(networkContract.methods.setInfo(srcString2, tokens[i]));
+    await sendTx(networkContract.methods.setInfo(destString2, ethAddress));
   }
```
ilanDoron: I am not sure why you are adding this.
My script does query which token listed per reserve in the network contract.

ilanDoron: I assume set info required also for network state - for the UI

contracts/KyberNetwork.sol
yaronvel: Network state will be listed as `info(sha3("network state")), or any other way. It is in application level, not contract level.

web3deployment/deployer.js
yaronvel: Im token asked for it.

yaronvel: One we define it, yes.

yaronvel:
This is something to settle with andrew and victor. For now, we ignore it.

PR #105, opened 2018-02-04
bound step functions to 10 steps

yaronvel:


yaronvel:
I didn't try to optimize it. just set everything to 6M.
Truffle sucks anyway, with web3 it takes < 3M gas.

PR #111, opened 2018-02-07
improve web3deployment/deployer.js 

Xcrux:
for example, it can now be used as follows:
```
./web3deployment/deployer.js --config-path ./web3deployment/deployment_script_input_kovan.json --gas-price-gwei 50 --print-private-key --rpc-url http://localhost:8545 --signed-tx-output ./txs.json
```

1. you can rebroadcast  txs.json later if there is an issue
2. you can print the private key if you feel comfortable or if in an offline mode

basically, you can remove nonce and replace it with `signedTxs.length`, but it doesn't really matter 

```
Usage: deployer.js --config-path [path] --gas-price-gwei [gwei]
--print-private-key [bool] --rpc-url [url] --signed-tx-output [path]

Options:
  --help          Show help                                            [boolean]
  --version       Show version number                                  [boolean]
  --configPath                                                        [required]
  --gasPriceGwei                                                      [required]
  --rpcUrl                                                            [required]
```


Xcrux:
I added an option to run without sending transaction:

you can (optionally) even run on localhost:8545 with a specified chain Id (mainly for estimating gas)

```
./web3deployment/deployer.js --config-path ./web3deployment/deployment_script_input_kovan.json --gas-price-gwei 1 --print-private-key --rpc-url http://localhost:8545 --signed-tx-output ./txs.json --dont-send-tx --chain-id 42
```
```
./web3deployment/sendTxs.js --rpc-url https://kovan.infura.io --signedTxsInput ./txs.json
```

it's also good to add chainId in the beginning (through `web3.eth.net.getId()`) because otherwise web3 will do it for every transaction (which can be a little slow)

PR #113, opened 2018-02-07
Yaronvel/mock step function setter

yaronvel:
add mock operator


yaronvel:
this is temp solution until core system will provide GUI for that.

PR #114, opened 2018-02-07
Ilan/auto deploy full output

ilanDoron:
my PR is adding the relevant tokens to develop
is that ok?


ilanDoron:
making a clean PR

PR #118, opened 2018-02-08
fix expected rate event: set quantity factor

ilanDoron:
unfortunatly
order of parameters was bad.


ilanDoron:
These commits are handled in commit #246

PR #119, opened 2018-02-08
require had wrong argument. one require has no meaning. doulbe logic

ilanDoron:


ilanDoron:
issue handled in another PR

PR #145, opened 2018-02-26
Ilan/print parameter for spyros

ilanDoron:
printing paramters for spyros.
will not print by default.
also printing knc wallet balance.


web3deployment/readVerifyDeployment.js
```diff
@@ -1329,4 +1390,4 @@ const tokensJson = {
           "address": "0x744d70fdbe2ba4cf95131626614a1763df805b9e",
         }
     }
-};
+};9
```
yaronvel: what is this 9?

ilanDoron: its a mistake.

PR #149, opened 2018-03-06
wrapper for fee burner + tests

ilanDoron:


ilanDoron:
keep working on fee burner wrapper

PR #150, opened 2018-03-08
Ilan/verify script with wrapper

ilanDoron:
verification script will verify code of wrap conversion rate if its address is set in json.

also some small improvments in script.

also fix addresses of admin and operators in staging json.


ilanDoron:
changes for readVerifyDeployement are included in another PR
will open PR for changes in staging json.

PR #161, opened 2018-03-18
Ilan/fix trades per day script

ilanDoron:
in the script I neglected all decimals.
now with decimals. calculation is more accurate.


ilanDoron:
included in another PR

PR #162, opened 2018-03-19
 add mock contract to simulate DGX Dex.

talbaneth:


contracts/mockContracts/MockDGXDex.sol
```diff
@@ -0,0 +1,187 @@
...
+import "./ecverify.sol";
+
+
+/// @title Mock Digix DEX
+/// @author Tal Baneth
+/// @dev a dummy contract that simulates Digix contract for purchase/sell with verified signatures. 
+contract MockDGXDEX is Withdrawable {
+    uint  constant internal G_TO_MG = (10**3);
+    uint  constant internal DGX_DECIMALS = (10**9);
+
```
ilanDoron: we don't use this notation.
only camel back notations...
Dabashot :)s

contracts/mockContracts/MockDGXDex.sol
```diff
@@ -0,0 +1,187 @@
...
+import "../ERC20Interface.sol";
+import "../Withdrawable.sol";
+import "./ecverify.sol";
+
+
+/// @title Mock Digix DEX
+/// @author Tal Baneth
+/// @dev a dummy contract that simulates Digix contract for purchase/sell with verified signatures. 
+contract MockDGXDEX is Withdrawable {
+    uint  constant internal G_TO_MG = (10**3);
```
ilanDoron: can u add some documentation what this means?
or maybe can be clearer names?

contracts/mockContracts/MockDGXDex.sol
```diff
@@ -0,0 +1,187 @@
...
+            verify_signed_price(block_number,
+                                nonce,
+                                weiPerDgxMg,
+                                signer,
+                                signature);
+
+            require(verified);
+        }
+        if(blockCheck) {
+            require((block_number + 5) >= block.number);
```
ilanDoron: this value - 5 - better be configurable...
anyway would rather not use numbers but  const values.d

contracts/mockContracts/MockDGXDex.sol
```diff
@@ -0,0 +1,187 @@
...
+        require((block_number + 5) >= block.number);
+
+        tokenAmount = (msg.value / weiPerDgxMg) * ( DGX_DECIMALS / G_TO_MG);
+        require(token.transfer(msg.sender, tokenAmount));
+
+        success = true;
+        purchasedAmount = tokenAmount;
+        
+        Purchase(success, purchasedAmount);
+
```
ilanDoron: seems like success always true here.
would add to this event more data: buyer address, price
or are u copying digix contract?

contracts/mockContracts/MockDGXDex.sol
```diff
@@ -0,0 +1,187 @@
...
+        success = true;
+        purchasedAmount = tokenAmount;
+        
+        Purchase(success, purchasedAmount);
+        
+    }
+
+    function sell(uint256 amount,
+                  uint256 block_number,
```
ilanDoron: notation.

contracts/mockContracts/MockDGXDex.sol
```diff
@@ -0,0 +1,187 @@
...
+                  uint256 weiPerDgxMg,
+                  address signer,
+                  bytes signature) public returns (bool success) {
+
+        uint256 amountWei;
+
+        if (sigCheck) {
+            bool verified;
+            address actual_signer;
+
```
ilanDoron: notation

contracts/mockContracts/MockDGXDex.sol
```diff
@@ -0,0 +1,187 @@
...
+                                signature);
+
+            require(verified);
+        }
+        if (blockCheck) {
+            require((block_number + 5) >= block.number);
+        }
+
+        amountWei = amount * G_TO_MG * weiPerDgxMg / DGX_DECIMALS;
+        require(token.transferFrom(msg.sender, this, amount));
```
ilanDoron: can't verify this formula. not sure what each number means.

contracts/mockContracts/MockDGXDex.sol
```diff
@@ -0,0 +1,187 @@
...
+        
+        success = true;
+
+        Sell(success, amountWei);
+    }
+
+    function verify_signed_price(uint _block_number, uint _nonce, uint _price, address _signer, bytes _signature)
+        public
+        returns (bool _verified, address _actual_signer)
+        {
```
ilanDoron: { should be indented left.
Also all next lines of this function.

contracts/mockContracts/MockDGXDex.sol
```diff
@@ -0,0 +1,187 @@
...
+        msg.sender.transfer(amountWei);
+        
+        success = true;
+
+        Sell(success, amountWei);
+    }
+
+    function verify_signed_price(uint _block_number, uint _nonce, uint _price, address _signer, bytes _signature)
+        public
+        returns (bool _verified, address _actual_signer)
```
ilanDoron: why do u start here with underscore?

contracts/mockContracts/MockDGXDex.sol
```diff
@@ -0,0 +1,187 @@
+pragma solidity ^0.4.18;
+
+//ask:
+//dst address change?
+//should init another token outside in test?
+
```
ilanDoron: yes.
we have testToken contract for creating a token.
u can see it is done in many tests...
just init it and name it digix.

contracts/mockContracts/MockDGXDex.sol
```diff
@@ -0,0 +1,187 @@
...
+// blocknumber + 5 >= current block number ? revert
+// signatures - skip for now.
```
ilanDoron: require will revert. this if fine.

test/mockDGXDEX.js
```diff
@@ -0,0 +1,134 @@
...
+        
+        // create dex with the given token. 
+        dgxDex = await mockDGXDEX.new(dgxToken.address, 0, accounts[0]);
+
+        // transfer 100 tokens to dex.
+        let tokenWei = (new BigNumber(10).pow(11)); // 10^9 is one
+        await dgxToken.transfer(dgxDex.address, tokenWei);
+ 
+        // store 10 eth in the dex.
+        let etherWei = new BigNumber(10).pow(19); // 10^18 is one
```
ilanDoron: pow should be 18

contracts/mockContracts/MockDGXDex.sol
```diff
@@ -0,0 +1,187 @@
...
+    function setBlockCheck (bool _blockCheck) public onlyAdmin {
+        blockCheck = _blockCheck;
+    }
+
+    function purchase(uint256 block_number,
+                      uint256 nonce,
+                      uint256 weiPerDgxMg,
+                      address signer,
+                      bytes signature) payable public returns (bool success, uint256 purchasedAmount) {
+
```
ilanDoron: here the { should move to next line.
also payable public ...
if { is not in function line it should move to next line and indented with "function" column

contracts/mockContracts/MockDGXDex.sol
talbaneth: I do see you use this notation for constants.

contracts/mockContracts/MockDGXDex.sol
talbaneth: changed to GRAM_TO_MILLIGRAM.

contracts/mockContracts/MockDGXDex.sol
talbaneth: Done. Added a const.

contracts/mockContracts/MockDGXDex.sol
talbaneth: The success return (always true) is copied from their contract.
I added more info to the event.

contracts/mockContracts/MockDGXDex.sol
talbaneth: doen. (the var names were copied from digix, but I changed them to camelcase notation now).

contracts/mockContracts/MockDGXDex.sol
talbaneth: done.

test/mockDGXDEX.js
```diff
@@ -0,0 +1,134 @@
+var TestToken = artifacts.require("./mockContracts/TestToken.sol");
+var mockDGXDEX = artifacts.require("./mockContracts/MockDGXDEX.sol")
+var Helper = require("./helper.js")
+
+var myToken;
```
ilanDoron: myToken isn't used...

test/mockDGXDEX.js
```diff
@@ -0,0 +1,134 @@
...
+
+var ethAddress = '0x00eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee';
+const BigNumber = require('bignumber.js');
+
+contract('MockDGXDEX', function (accounts) {
+    it("create DGX dex and store funds in it.", async function (){
+
+        // create dgx token.
+        dgxToken = await TestToken.new("dex dgx token", "dgx", 9);
+
```
ilanDoron: I think the correct way is to declare this in global scope. and all other globals.
using let.
I see its used during test...

contracts/mockContracts/MockDGXDex.sol
talbaneth: Changed the calculation to be in steps (also on purchase function). Hope it's clearer now.

Purchase:
amountMg = (msg.value / weiPerDgxMg)
amountMgDgxWei = amountMg * DGX_DECIMALS;
amountDgxWei = amountMgDgxWei / GRAM

Sell:
amountDgxWei = amount;
amountMgDgxWei = amountDgxWei * GRAM_TO_MILLIGRAM;
amountWei = amountMgDgxWei * weiPerDgxMg / DGX_DECIMALS;

contracts/mockContracts/MockDGXDex.sol
talbaneth: done.

contracts/mockContracts/MockDGXDex.sol
talbaneth: To distinguish it from storage variables that have the same nam.
I Saw it done in the rest of your contracts.

contracts/mockContracts/MockDGXDex.sol
talbaneth: done. removed the comments.

contracts/mockContracts/MockDGXDex.sol
talbaneth: done.

talbaneth:
handled the issues you mentioned.
I did leave the constant notations and the _<var> notation when passing argument with same name as a storage variable. This is same as other contracts.

PR #163, opened 2018-03-19
Ilan/update staging json

ilanDoron:
sanity rates new contract + operator.

fee burner - change admin to wrapper.
update max per block and max total for 3 coins.


contracts/DigixConnector.sol
```diff
@@ -0,0 +1,7 @@
+pragma solidity ^0.4.0;
```
yaronvel: is this file relevant to the PR? please remove it

ilanDoron:
done.
removed files.

PR #169, opened 2018-03-28
Ilan/support new tokens

ilanDoron:


ilanDoron:
many changes to avoid un necessary alerts (red color) from verification script.

PR #177, opened 2018-04-11
Fee handler master

talbaneth:


talbaneth:
Closing this. 
Will make a RP to a develop based branch so we can see better diffs.

PR #180, opened 2018-04-15
add DAI token, add internal use flag

yaronvel:


ilanDoron:
this pull request not relevant.
all  data added in PR 181
internal use flag for staging and DAI for staging are both added in PR 181

PR #184, opened 2018-04-20
enable token to token trade on kyber network

ilanDoron:
please don't pull
only added this PR so its easier to see differences


contracts/KyberNetwork.sol
```diff
@@ -40,8 +40,8 @@ contract KyberNetwork is Withdrawable, Utils {
     }
     /* solhint-enable no-complex-fallback */
 
-    event ExecuteTrade(address indexed sender, ERC20 src, ERC20 dest, uint actualSrcAmount, uint actualDestAmount);
-
+    event expected(uint actualDest, uint expectedDest);
```
yaronvel: never used. and bad convention name.

contracts/KyberNetwork.sol
```diff
@@ -90,8 +90,8 @@ contract KyberNetwork is Withdrawable, Utils {
         userSrcBalanceAfter = getBalance(src, msg.sender);
         userDestBalanceAfter = getBalance(dest, destAddress);
 
-        require(userSrcBalanceAfter <= userSrcBalanceBefore);
-        require(userDestBalanceAfter >= userDestBalanceBefore);
+        require(userSrcBalanceAfter < userSrcBalanceBefore);
```
yaronvel: why strict inequality?

contracts/KyberNetwork.sol
```diff
@@ -214,11 +214,39 @@ contract KyberNetwork is Withdrawable, Utils {
     /// @dev best conversion rate for a pair of tokens, if number of reserves have small differences. randomize
     /// @param src Src token
     /// @param dest Destination token
+    function findBestRate(ERC20 src, ERC20 dest, uint srcQty) public view returns(uint noUse, uint rate) {
```
yaronvel: why return param is called `noUse`?

contracts/KyberNetwork.sol
```diff
@@ -214,11 +214,39 @@ contract KyberNetwork is Withdrawable, Utils {
...
+
+        (reserve2, rateEthToDest) = searchBestRate(ETH_TOKEN_ADDRESS, dest, ethAmount);
+        actualDest = calcDestAmount(ETH_TOKEN_ADDRESS, dest, ethAmount, rateEthToDest);
+
+        if (rateSrcToEth == PRECISION) rate = rateEthToDest;
+        else if (rateEthToDest == PRECISION) rate = rateSrcToEth;
+        else rate = calcRateFromQty(srcQty, actualDest, getDecimals(src), getDecimals(dest));
+    }
+
     /* solhint-disable code-complexity */
```
yaronvel: you can probably remove it now right?

contracts/KyberNetwork.sol
```diff
@@ -356,6 +403,13 @@ contract KyberNetwork is Withdrawable, Utils {
     {
         uint callValue = 0;
 
+        if (src == dest) {
+            //this is for a "fake" trade when both src and dest are ethers.
+            if (destAddress != (address(this)))
```
yaronvel: is this good? looks bad.

contracts/KyberNetwork.sol
ilanDoron: because it can't remain the same. if its same it means no trade was done. better revert.

PR #194, opened 2018-05-08
update verification script to enable fast fetch of step functions. use service contract. show liquidity conversion rate

ilanDoron:
updating mainnet json.
also adding some missing addresses to permission.

a faster way to fetch step function, using some service contract.


ilanDoron:
this was handled in another PR

PR #200, opened 2018-05-09
delete a srcQty require

rotcivegaf:
this require is unnecessary
        require(srcQty <= MAX_QTY);
in the next line check this and previus check quantityFactor is not zero.


ilanDoron:
notice
if srcQty > MAX_QTY and srcQty is very high.
srcQty * quantityFactor could overflow and then 2nd require 
require(srcQty * quantityFactor <= MAX_QTY); 
will not fail

That's why we left both.

PR #201, opened 2018-05-09
the variable bestReserve is never used

rotcivegaf:


yaronvel:
Thanks for the PR, but linter give warnings when using such a syntax.

PR #204, opened 2018-05-15
don't respect the ERC20 standard

rotcivegaf:
this functions should will return a bool:
    function transferFrom(address from, address to, uint value) public returns(bool);
    function approve(address spender, uint value) public returns (bool success);
    function transfer(address to, uint value) public;

for example on KyberNetwork.sol in line 372, need that

### DIFF
--- a/contracts/mockContracts/TestToken.sol
+++ b/contracts/mockContracts/TestToken.sol
@@ -197,4 +197,15 @@ contract TestToken is StandardToken {
         transferFrom( _from, msg.sender, _value );
         return burn(_value);
     }
+    
+    //
+    // WARNING!!! use to create tokens ONLY in test cases
+    //
+    function createTokens(address beneficiary, uint256 tokens) public returns (bool) {
+        balances[beneficiary] = tokens.add(balances[beneficiary]);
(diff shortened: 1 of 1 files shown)


ilanDoron:
Thank u for this PR

Actually this non compliance in test token helped us to find a similar bug in OMG token.
But I will merge this.
just one thing missing
the prototype for approve wasn't updated in interface file...
@rotcivegaf would like to update it as well?

contracts/mockContracts/TestToken.sol
```diff
@@ -102,19 +102,20 @@ contract BasicToken is ERC20Basic {
...
     modifier onlyPayloadSize(uint size) {
         if (msg.data.length < size + 4) {
-         revert();
+            revert();
         }
         _;
     }
 
-    function transfer(address _to, uint _value)  public onlyPayloadSize(2 * 32) {
+    function transfer(address _to, uint _value)  public onlyPayloadSize(2 * 32) returns(bool){
```
ilanDoron: notice prototype for transfer wasn't updated.
line 72 - above

ilanDoron:
This code was fixed in master branch.
so this code is not relevant anymore.
although suggested changes were valid.

PR #208, opened 2018-05-17
network v2 - add better decimals, enable set API per contract

ilanDoron:


ilanDoron:
this was handled in other PRs

PR #216, opened 2018-06-04
network proxy for kyber reserve + utils2

ilanDoron:


contracts/KyberNetwork.sol
```diff
@@ -450,17 +439,14 @@ contract KyberNetwork is Withdrawable, Utils {
         uint callValue = 0;
 
         if (src == dest) {
-            //this is for a "fake" trade when both src and dest are ethers.
+            //this is for a "fake" trade when both src and dest are ehters.
```
yaronvel: ehters? type

PR #237, opened 2018-06-21
fix deloy script for proxy

ilanDoron:
fix deployement.js script for running with proxy and changes in API


ilanDoron:
was PRed
already exists on develop branch

PR #245, opened 2018-06-26
Update staging json

talbaneth:


ilanDoron:
I think better add missing 3rd party reserve addresses.

talbaneth:
probably better to keep it identical to what Victor uses.

web3deployment/mainnet_staging.json
```diff
@@ -239,12 +239,96 @@
...
+      "listed": true
+    },
+    "BBO": {
+      "name": "Bigbom",
+      "decimals": 18,
+      "address": "0x84f7c44b6fed1080f647e354d552595be2cc602f",
+      "internal use": false,
+      "listed": true
+    },
     "ETH": {
```
ilanDoron: maybe keep Eth bottom token
as till now...

Issue #276, opened 2018-08-05
block recursive calls in KyberNetwork contract


yaronvel:
Failing to prevent that could allow malicious network admin to override the reserve step function mechanism when listing malicious tokens and performing token to token swap.

ilanDoron:
delivered to master in #373

PR #278, opened 2018-08-06
block re entrace in kyber trade.

ilanDoron:
for issue #276

### DIFF
--- a/contracts/KyberNetwork.sol
+++ b/contracts/KyberNetwork.sol
@@ -15,7 +15,8 @@ import "./FeeBurnerInterface.sol";
 /// @title Kyber Network main contract
 contract KyberNetwork is Withdrawable, Utils2, KyberNetworkInterface {
 
-    uint public negligibleRateDiff = 10; // basic rate steps will be in 0.01%
+    uint public negligibleRateDiff = 10; //
+    uint public entranceCounter = 1;       // to block function re-entrance
     KyberReserveInterface[] public reserves;
     mapping(address=>bool) public isReserve;
     WhiteListInterface public whiteListContract;
@@ -74,6 +75,8 @@ contract KyberNetwork is Withdrawable, Utils2, KyberNetworkInterface {
         require(hint.length == 0);
         require(msg.sender == kyberNetworkProxyContract);
 
+        uint localCounter = ++entranceCounter;
+
         TradeInput memory tradeInput;
 
         tradeInput.trader = trader;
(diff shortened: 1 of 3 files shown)


contracts/KyberNetwork.sol
```diff
@@ -15,7 +15,8 @@ import "./FeeBurnerInterface.sol";
 /// @title Kyber Network main contract
 contract KyberNetwork is Withdrawable, Utils2, KyberNetworkInterface {
 
-    uint public negligibleRateDiff = 10; // basic rate steps will be in 0.01%
+    uint public negligibleRateDiff = 10; //
```
yaronvel: why the change here?

contracts/KyberNetwork.sol
```diff
@@ -374,7 +381,7 @@ contract KyberNetwork is Withdrawable, Utils2, KyberNetworkInterface {
         require(validateTradeInput(tradeInput.src, tradeInput.srcAmount, tradeInput.dest, tradeInput.destAddress));
 
         BestRateResult memory rateResult =
-        findBestRateTokenToToken(tradeInput.src, tradeInput.dest, tradeInput.srcAmount);
+            findBestRateTokenToToken(tradeInput.src, tradeInput.dest, tradeInput.srcAmount);
```
yaronvel: why change?

contracts/KyberNetwork.sol
```diff
@@ -86,7 +89,11 @@ contract KyberNetwork is Withdrawable, Utils2, KyberNetworkInterface {
         tradeInput.walletId = walletId;
         tradeInput.hint = hint;
 
-        return trade(tradeInput);
+        uint actualDestAmount = trade(tradeInput);
```
yaronvel: I think modifier will be nicer solution. need to read about it though. I don't think return value is an issue.

contracts/KyberNetwork.sol
ilanDoron: it should be indented.
really

yaronvel: it is not time for such changes

contracts/KyberNetworkInterface.sol
```diff
@@ -16,5 +16,5 @@ interface KyberNetworkInterface {
         returns (uint expectedRate, uint slippageRate);
 
     function tradeWithHint(address trader, ERC20 src, uint srcAmount, ERC20 dest, address destAddress,
-        uint maxDestAmount, uint minConversionRate, address walletId, bytes hint) public payable returns(uint);
+        uint maxDestAmount, uint minConversionRate, address walletId, bytes hint) external payable returns(uint);
```
yaronvel: please revert. it will change bytecode.

ilanDoron: but byte code will change anyway since we apply this reentrant guard.

PR #280, opened 2018-08-12
Permission less reserve

yaronvel:
just to make comments


contracts/PermissionLessReserve.sol
```diff
@@ -0,0 +1,541 @@
...
+        require(burner != address(0));
+
+        kncToken.approve(feeBurnerContract, 0);
+
+        feeBurnerContract = burner;
+
+        kncToken.approve(feeBurnerContract, (2**255));
+    }
+
+    function isOrderBetterRate(uint32 orderID, uint32 checkedOrderID) public view returns(bool) {
```
yaronvel: I stopped reading here. will return to it later.

PR #288, opened 2018-08-26
Orders external instances

amitkot:


amitkot:
@ilanDoron

contracts/Orders.sol
```diff
@@ -1,56 +1,31 @@
...
 
     mapping (uint32 => Order) public orders;
 
-    uint8 constant public ETH_TO_TOKEN = 1;
-    uint8 constant public TOKEN_TO_ETH = 2;
-
-    uint32 constant public TAIL_ID = 1;
-    uint32 constant public BUY_HEAD_ID = 2;
-    uint32 constant public SELL_HEAD_ID = 3;
+    uint32 constant public TAIL_ID = 0;
```
ilanDoron: don't like using 0. unless u r sure its best practice.
since all is init by default to 0. if we skip some init we think it points on tail when we don't mean it!!!

contracts/Orders.sol
```diff
@@ -1,56 +1,31 @@
 pragma solidity 0.4.18;
 
-
 import "./Utils2.sol";
+import "./Withdrawable.sol";
```
ilanDoron: line space

contracts/Orders.sol
```diff
@@ -1,56 +1,31 @@
 pragma solidity 0.4.18;
 
-
 import "./Utils2.sol";
```
ilanDoron: line space above

2 lines required

contracts/Orders.sol
```diff
@@ -1,56 +1,31 @@
...
+    uint32 constant public TAIL_ID = 0;
+    uint32 constant public HEAD_ID = 1;
 
-    uint32 public nextId = 4;
-
-    Order internal BUY_HEAD;
-    Order internal SELL_HEAD;
+    uint32 public nextId = 2;
 
     function Orders() public {
```
ilanDoron: constructor is a good place to set admin...

contracts/Orders.sol
ilanDoron: @amitkot 
not sure.
I am sure its needed before contract.
but in all other contract its also before the imports. so consider doing it anyway...

contracts/Orders.sol
amitkot: Done.

PR #290, opened 2018-08-27
Moving some functionality to MakerOrders contract

amitkot:
The responsibility for keeping track of each makers free orders is
moved to a new layer that the PermissionLessReserve now extends.

### DIFF
--- a/contracts/MakerOrders.sol
+++ b/contracts/MakerOrders.sol
@@ -1,0 +1,79 @@
+pragma solidity 0.4.18;
+
+
+import "./Withdrawable.sol";
+
+
+contract MakerOrders is Withdrawable {
+    struct FreeOrders {
+        uint32 firstOrderId;
+        uint32 numOrders; //max is 256
+        uint8 maxOrdersReached;
+        uint256 takenBitmap;
+    }
+
+    //each maker will have orders that will be reused.
+    mapping(address => FreeOrders) public makerOrdersSell;
+    mapping(address => FreeOrders) public makerOrdersBuy;
+
+    function takeOrderId(FreeOrders storage freeOrders)
+        internal
+        onlyAdmin
+        returns(uint32)
+    {
+        uint numOrders = freeOrders.numOrders;
+        uint orderBitmap = freeOrders.takenBitmap;
+        uint bitPointer = 1;
+
+        for(uint i = 0; i < numOrders; ++i) {
+            if ((orderBitmap & bitPointer) == 0) {
+                freeOrders.takenBitmap = orderBitmap | bitPointer;
+                return(uint32(uint(freeOrders.firstOrderId) + i));
+            }
+
+            bitPointer *= 2;
+        }
+
+        freeOrders.maxOrdersReached = 1;
+
+        require(false);
+    }
+
+    /// @dev mark order as free to use.
+    function releaseOrderId(FreeOrders storage freeOrders, uint32 orderId)
+        internal
+        onlyAdmin
+        returns(bool)
+    {
+        require(orderId >= freeOrders.firstOrderId);
+        require(orderId < (freeOrders.firstOrderId + freeOrders.numOrders));
+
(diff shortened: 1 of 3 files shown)


amitkot:
@ilanDoron

contracts/MakerOrders.sol
```diff
@@ -0,0 +1,79 @@
...
+    struct FreeOrders {
+        uint32 firstOrderId;
+        uint32 numOrders; //max is 256
+        uint8 maxOrdersReached;
+        uint256 takenBitmap;
+    }
+
+    //each maker will have orders that will be reused.
+    mapping(address => FreeOrders) public makerOrdersSell;
+    mapping(address => FreeOrders) public makerOrdersBuy;
```
ilanDoron: thought the mapping should be in extending contract. its not used here anyway. and this layer don't have to be aware of sell buy diff.

thats what we talked...

contracts/MakerOrders.sol
```diff
@@ -0,0 +1,79 @@
...
+
+    function takeOrderId(FreeOrders storage freeOrders)
+        internal
+        onlyAdmin
```
ilanDoron: since internal and extended. don't need only Admin no more...

contracts/MakerOrders.sol
```diff
@@ -0,0 +1,79 @@
...
+                freeOrders.takenBitmap = orderBitmap | bitPointer;
+                return(uint32(uint(freeOrders.firstOrderId) + i));
+            }
+
+            bitPointer *= 2;
+        }
+
+        freeOrders.maxOrdersReached = 1;
+
+        require(false);
```
ilanDoron: should decide if to use maxOrdersReached and avoid reverting.
or revert.

if we revert. maxOrdersReached will not be set anyway.

contracts/MakerOrders.sol
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    /// @dev mark order as free to use.
+    function releaseOrderId(FreeOrders storage freeOrders, uint32 orderId)
+        internal
+        onlyAdmin
```
ilanDoron: again. since internal don't need onlyAdmin

contracts/MakerOrders.sol
```diff
@@ -0,0 +1,79 @@
...
+        freeOrders.takenBitmap &= bitNegation;
+    }
+
+    function allocateOrders(
+        FreeOrders storage freeOrders,
+        uint32 firstAllocatedId,
+        uint32 howMany
+    )
+        internal
+        onlyAdmin
```
ilanDoron: remove onlyAdmin

contracts/Orders.sol
```diff
@@ -260,73 +260,12 @@ contract Orders is Withdrawable, Utils2 {
         );
     }
 
-    function getNextOrder(uint32 orderId) public view returns(uint32, bool isLast) {
+    function getNextOrder(uint32 orderId)
+        public
+        view
+        returns(uint32, bool isLast)
+    {
```
ilanDoron: I prefer the one line syntax if possible
Don't u?

contracts/MakerOrders.sol
amitkot: Done.

contracts/MakerOrders.sol
amitkot: Done.

contracts/MakerOrders.sol
amitkot: Done.

contracts/Orders.sol
amitkot: Following the Solidity style guide, if it exceeds the line length we should break it.

contracts/MakerOrders.sol
amitkot: Done.

contracts/MakerOrders.sol
amitkot: Let's discuss this.

PR #293, opened 2018-08-29
Permissionless Orders: Moving things around

amitkot:


amitkot:
@ilanDoron

contracts/permissionless/Orders.sol
```diff
@@ -62,6 +62,7 @@ contract Orders is Withdrawable, Utils2 {
     )
         public
         onlyAdmin
+        // TODO: do not return value
         returns(uint32)
```
ilanDoron: would be nice to document what do u return.
or at least name it...

PR #296, opened 2018-09-02
Moving functions to Reserve level and fix for Orders

amitkot:


amitkot:
@ilanDoron

contracts/permissionless/OrderBookReserve.sol
```diff
@@ -738,4 +738,30 @@ contract OrderBookReserve is MakerOrders, Utils2, KyberReserveInterface {
         require(weiAmount > 0);
         return weiAmount;
     }
+
+    function getOrderData(uint32 orderId) public view
```
ilanDoron: u never compiled????

ilanDoron: this is not close to the prototype u use...
:(

PR #298, opened 2018-09-03
Fix for breaking OrderBookReserve change + Orders add changes

amitkot:
- Add after id now returns false on bad hints, rather than revert.


amitkot:
@ilanDoron conflict fixed

PR #299, opened 2018-09-03
Update with hint returns false for bad hints

amitkot:


amitkot:
@ilanDoron currently working on the optimisation for only-update-amounts. Will arrive in the next PR.

PR #302, opened 2018-09-06
Orders: improving updateWithPositionHint to updateAmounts in case no move required

amitkot:
Also some test ordering and renaming.

### DIFF
--- a/contracts/permissionless/Orders.sol
+++ b/contracts/permissionless/Orders.sol
@@ -21,6 +21,11 @@ contract Orders is Withdrawable, Utils2, OrdersInterface {
     uint32 constant public TAIL_ID = 1;
     uint32 constant public HEAD_ID = 2;
 
+    // Results of calling updateWithPositionHint.
+    uint8 constant public UPDATE_ONLY_AMOUNTS = 0;
+    uint8 constant public UPDATE_MOVE_ORDER = 1;
+    uint8 constant public UPDATE_FAILED = 2;
+
     uint32 public nextFreeId = 3;
 
     function Orders(address _admin) public {
@@ -64,6 +69,8 @@ contract Orders is Withdrawable, Utils2, OrdersInterface {
         public
         onlyAdmin
     {
+        require(orderId != 0 && orderId != HEAD_ID && orderId != TAIL_ID);
+
         uint32 prevId = findPrevOrderId(srcAmount, dstAmount);
         addAfterValidId(maker, orderId, srcAmount, dstAmount, prevId);
     }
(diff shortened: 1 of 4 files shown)


amitkot:
@ilanDoron please review.

contracts/permissionless/Orders.sol
```diff
@@ -138,6 +126,46 @@ contract Orders is Withdrawable, Utils2, OrdersInterface {
...
+    function updateAmounts(
+        uint32 orderId,
+        uint128 srcAmount,
+        uint128 dstAmount
+    )
+        public
+        onlyAdmin
+        returns (bool)
+    {
+        require(orderId != HEAD_ID && orderId != TAIL_ID);
```
ilanDoron: code for this function is a copy of the code in isRightPosition

if isRightPosition will have another parameter (nextId) u will be able to use it for this.

contracts/permissionless/Orders.sol
```diff
@@ -116,16 +114,6 @@ contract Orders is Withdrawable, Utils2, OrdersInterface {
...
-        bool hintIsRightPositionAfterUpdate = false;
-        if (hintIsCurrentPosition && hintIsRightPositionAfterUpdate) {
-            // Order is in the right position, update amounts
-            orders[orderId].srcAmount = srcAmount;
-            orders[orderId].dstAmount = dstAmount;
-            AmountUpdateOnly();
-            return true;
-        }
-
         if (isRightPosition(srcAmount, dstAmount, prevId)) {
```
ilanDoron: if the order doesn't change position here the test might fail while still the correct position.
I know in our design it shouldn't happen. but still. its a flow in the code.
I can build a test that will prove it....

ilanDoron: I mean. the correct test will make this update call fail...

contracts/permissionless/Orders.sol
amitkot: updateAmounts merged into updateWithPositionHint, using a single validation function as we discussed f2f.

contracts/permissionless/Orders.sol
amitkot: Agree, added test and fixed the issue.

amitkot:
@ilanDoron merged updateAmounts to udpateWithHint. Please review and merge.

contracts/permissionless/Orders.sol
```diff
@@ -213,38 +226,41 @@ contract Orders is Withdrawable, Utils2, OrdersInterface {
...
+        uint32 prevId,
+        uint32 nextId
     )
         private
         view
         returns (bool)
     {
-        // Make sure prev is not the tail.
-        if (prevId == TAIL_ID) return false;
+        if (prevId == TAIL_ID || nextId == HEAD_ID) return false;
```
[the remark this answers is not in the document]
ilanDoron: sorry. its testing another thing. my error...

PR #319, opened 2018-10-14
create mock ENS the imitates ENS structure

ilanDoron:


ilanDoron:
Mock ENS doesn't need all these changes.

PR #321, opened 2018-10-17
Add get ERC20 tokens allowances for wrapper

manhlx3006:


tranvictor:
@yaronvel @ilanDoron hi everyone, we want to improve the performance of the trading API because we have to process an API to get `token enabled` information of an address. This PR adds a function for that. Please check the PR.

PR #338, opened 2018-11-27
order book reserve PR for review only. don't pull

ilanDoron:
includes order book resreve after fixing stake model and fixing some more comments from Tal


ilanDoron:
this was only for code review.

PR #347, opened 2018-12-19
Ilan/fix free knc calculation

ilanDoron:


ilanDoron:
will add this code in another PR

PR #354, opened 2018-12-31
orderbook coverage

ilanDoron:


contracts/permissionless/OrderbookReserve.sol
```diff
@@ -20,7 +20,7 @@ interface MedianizerInterface {
 
 contract OrderbookReserve is OrderIdManager, Utils2, KyberReserveInterface, OrderbookReserveInterface {
 
-    uint public constant BURN_TO_STAKE_FACTOR = 4;      // stake per order must be x4 then expected burn amount.
+    uint public constant BURN_TO_STAKE_FACTOR = 5;      // stake per order must be x4 then expected burn amount.
```
talbaneth: comment needs to change to 5 as well.

contracts/KyberNetworkInterface.sol
```diff
@@ -12,9 +12,6 @@ interface KyberNetworkInterface {
     function enabled() public view returns(bool);
     function info(bytes32 id) public view returns(uint);
 
-    function searchBestRate(ERC20 src, ERC20 dest, uint srcAmount, bool usePermissionless) public view
```
talbaneth: was this removed because it shouldn't be used externally? (since we have getExpectedRate)

ilanDoron: its removed since its not used. not because it shouldn't be used.

PR #358, opened 2019-01-07
OrderList: mark and detect deleted orders

amitkot:
Indicate deleted orders by setting their prev to TAIL and their next to 
HEAD.

Use this indication to refuse accepting them as hints.


ilanDoron:
please make sure to test when empty list and bad hint ( a deleted order)

PR #365, opened 2019-01-08
Ilan/updgrade verification script v3

ilanDoron:


ilanDoron:
will be done in anther PR

PR #371, opened 2019-01-14
fix bug that updating canceled order can revive order

ilanDoron:


contracts/permissionless/OrderList.sol
```diff
@@ -281,7 +281,7 @@ contract OrderList is PermissionGroups, OrderListInterface {
         Order storage order = orders[orderId];
 
         // Make sure such order exists in mapping.
-        require(order.prevId != 0 || order.nextId != 0);
+        require(order.prevId != 0 && order.nextId != 0 && order.prevId != TAIL_ID && order.nextId != HEAD_ID);
```
amitkot: Please also add tests to `orderlist.js` to check these different requirements directly.

amitkot: Consider splitting the two situations: 
- Order was never in the list (order.prevId == 0 or order.nextId == 0), and 
- Order was already removed (prevId == TAIL or nextId == HEAD)

test/orderbookReserve.js
```diff
@@ -3193,6 +3193,62 @@ contract('OrderbookReserve', async (accounts) => {
             assert.equal(orderList.length, 4);
         })
 
+        it("test can't 'revive' canceled (tok to eth) order by updating it", async() => {
+            let tokenWeiDepositAmount = new BigNumber(500 * 10 ** 18);
+            let kncTweiDepositAmount = 600 * 10 ** 18;
+            let ethWeiDepositAmount = (new BigNumber(0 * 10 ** 18)).add(30000);
```
amitkot: Maybe I'm missing something here, can you replace with
```js
let ethWeiDepositAmount = new BigNumber(30000);
```
?

test/orderbookReserve.js
```diff
@@ -3193,6 +3193,62 @@ contract('OrderbookReserve', async (accounts) => {
...
+            await makerDeposit(maker1, ethWeiDepositAmount, tokenWeiDepositAmount, kncTweiDepositAmount);
+
+            let orderSrcAmountTwei = new BigNumber(9 * 10 ** 18);
+            let orderDstWei = new BigNumber(2 * 10 ** 18);
+
+            await reserve.submitTokenToEthOrder(orderSrcAmountTwei, orderDstWei, {from: maker1});
+            let rc = await reserve.submitTokenToEthOrder(orderSrcAmountTwei, orderDstWei.add(100), {from: maker1});
+            let id2ndOrder = rc.logs[0].args.orderId.valueOf();
+            rc = await reserve.submitTokenToEthOrder(orderSrcAmountTwei, orderDstWei.add(200), {from: maker1});
+            let id3rdOrder = rc.logs[0].args.orderId.valueOf();
```
amitkot: Nit: no need to save this id as a variable as you do not use it later.

test/orderbookReserve.js
```diff
@@ -3193,6 +3193,62 @@ contract('OrderbookReserve', async (accounts) => {
...
```
amitkot: Nit: no need to save this id as a variable as you do not use it later.

test/orderbookReserve.js
```diff
@@ -3193,6 +3193,62 @@ contract('OrderbookReserve', async (accounts) => {
             assert.equal(orderList.length, 4);
         })
 
+        it("test can't 'revive' canceled (tok to eth) order by updating it", async() => {
+            let tokenWeiDepositAmount = new BigNumber(500 * 10 ** 18);
```
amitkot: JS nit: it's probably better to use `const` rather than `let` if possible to minimize later assignments by mistake.

test/orderbookReserve.js
```diff
@@ -3193,6 +3193,62 @@ contract('OrderbookReserve', async (accounts) => {
...
+            await reserve.submitTokenToEthOrder(orderSrcAmountTwei, orderDstWei, {from: maker1});
+            let rc = await reserve.submitTokenToEthOrder(orderSrcAmountTwei, orderDstWei.add(100), {from: maker1});
+            let id2ndOrder = rc.logs[0].args.orderId.valueOf();
+            rc = await reserve.submitTokenToEthOrder(orderSrcAmountTwei, orderDstWei.add(200), {from: maker1});
+            let id3rdOrder = rc.logs[0].args.orderId.valueOf();
+            await reserve.submitTokenToEthOrder(orderSrcAmountTwei, orderDstWei.add(300), {from: maker1});
+
+            await reserve.cancelTokenToEthOrder(id2ndOrder, {from: maker1});
+
+            let list = await reserve.getTokenToEthOrderList();
```
amitkot: Nit: no need to save this id as a variable as you do not use it later.

contracts/permissionless/OrderList.sol
ilanDoron: will fix documentation here.
but not sure I can follow your strict conventions.

test/orderbookReserve.js
ilanDoron: I do use it. 
see below

PR #376, opened 2019-01-17
Uniswap

amitkot:


contracts/uniswap/UniswapReserve.sol
```diff
@@ -0,0 +1,300 @@
...
+            getDecimals(src), /* srcDecimals */
+            getDecimals(dest) /* dstDecimals */
+        );
+    }
+
+    event TradeExecute(
+        address indexed sender,
+        address src,
+        uint srcAmount,
+        address destToken,
```
ilanDoron: for src its src only.
for dest its destToken

better be the same
of is it different in Kyber classic reserve?

contracts/uniswap/UniswapReserve.sol
```diff
@@ -0,0 +1,300 @@
...
+        } else {
+            convertedQuantity = exchange.getTokenToEthInputPrice(srcQty);
+            convertedQuantity = convertedQuantity * (10000 - feeBps) / 10000;
+        }
+
+        return calcRateFromQty(
+            srcQty, /* srcAmount */
+            convertedQuantity, /* destAmount */
+            getDecimals(src), /* srcDecimals */
+            getDecimals(dest) /* dstDecimals */
```
ilanDoron: use dest in the comment
its dest in all other places (not dst).

contracts/uniswap/UniswapReserve.sol
```diff
@@ -0,0 +1,300 @@
...
+        ERC20 destToken,
+        address destAddress,
+        uint conversionRate,
+        bool validate
+    )
+        public
+        payable
+        returns(bool)
+    {
+        // Not using this variable that is part of the interface.
```
ilanDoron: there is a general question here.
is there an attack scenario where destAmount can change between getRate and trade?
this can happen due to any external contract call between the two.
was this option tested?

ilanDoron: in general the easiest defense for this is to verify rate is at least conversionRate in  parameter.

contracts/uniswap/UniswapReserve.sol
amitkot: I used the param name as it defined in utils2.sol:
https://github.com/KyberNetwork/smart-contracts/blob/master/contracts/Utils2.sol#L36

Issue #388, opened 2019-02-15
Automated Price Reserve Setup Fail


tphanson:
I met this kind of error when try to configuring Automated Price Reserve.
* Command:

```
node liquidityReserveDeployer.js --config-path ropsten.json --gas-price-gwei 30 --rpc-url https://ropsten.infura.io --print-private-key true --network-address "0x3f9a8e219ab1ad42f96b22c294e564b2b48fe636"
```

* Error details:

```
Error: Error: [number-to-bn] while converting number undefined to BN.js instance, error: invalid number value. Value must be an integer, hex string, BN or BigNumber instance. Note, decimals are not supported. Given value: "undefined"
    at Object.toBN (/Users/admin/Desktop/kyber-smart-contracts/node_modules/web3-utils/src/utils.js:64:15)
    at /Users/admin/Desktop/kyber-smart-contracts/web3deployment/liquidityReserveDeployer.js:159:46
    at Array.forEach (<anonymous>)
    at parseInput (/Users/admin/Desktop/kyber-smart-contracts/web3deployment/liquidityReserveDeployer.js:148:28)
    at Object.<anonymous> (/Users/admin/Desktop/kyber-smart-contracts/web3deployment/liquidityReserveDeployer.js:303:3)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:742:12)
    at startup (internal/bootstrap/node.js:282:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:743:3)
```

I invested a bit of my time and recognized that in [the guide](https://developer.kyber.network/docs/AutomatedReservesGuide/) you told that `The most important properties are symbol, name, decimals, and address. The rest can be ignored. So you can specify your token details, filling in these fields.` and indeed, you ignored them in ETH config.

```
"ETH": {
      "name": "Ethereum",
      "decimals": 18,
      "address": "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee"
    }
```

However, when I come to `liquidityReserveDeployer.js` I saw,

```
const dict = {
[...]

PR #389, opened 2019-02-18
Fixes to liquidityReserveDeployer.js script

ayobuenavista:
Fix for issue reported on #388.


ayobuenavista:
> thanks for the pr. since you already cleaning things, can you please also remove `"valid duration block"` from the json?

Removed the 'valid duration block' and the extra token in the JSON, since only one token definition is needed for Automated Price Reserve.

ayobuenavista:
> I see you also changed `OMG` to `KNC`.
> So please remove
> ` "OMG" : "0x44d34a119ba21a42167ff8b77a88f0fc7bb2db90",`
> from the withdraw addresses

OMG withdrawal address removed from JSON.

PR #396, opened 2019-03-07
handle zero qty on get conversion rate in permission less reserve

ilanDoron:


yaronvel:
please split to two prs.
one for expected rate
and one for permissionless, which is not deployed yet

PR #398, opened 2019-03-11
Revert "0 rate instead revert in get expected rate with rate > 1M"

ilanDoron:
Reverts KyberNetwork/smart-contracts#378


ilanDoron:
reverting this commit isn't required.
Since the code will be deployed

kawinduekanayaka:
Wow

PR #413, opened 2019-07-21
enable hinting which reserve to use. 1st version

ilanDoron:


yaronvel:
why asking for a PR as such early stage?

yaronvel:
Please only pr when ready

Issue #291, opened 2018-08-28
bad function call in conversion rate (pricing)


ilanDoron:
in conversion rate. 
https://github.com/KyberNetwork/smart-contracts/blob/02094d33d8d552adf8b4e58fbf38a3aabf27d12d/contracts/ConversionRates.sol#L227

call to getTokenQty(token, rate, qty);  
rate and qty are swapped
should be replaced by
getTokenQty(token, qty, rate);

manhlx3006:
Fixed with new conversion rate contract in this PR: https://github.com/KyberNetwork/smart-contracts/pull/416/files

manhlx3006:
This issue should be resolved in enhanced step function contract

PR #416, opened 2019-08-06
[WIP] Add new conversion rate contract with step function enhancement

manhlx3006:
It is working in progress for step function enhancement feature
Fixed: https://github.com/KyberNetwork/smart-contracts/issues/291

### DIFF
--- a/contracts/ConversionRates2.sol
+++ b/contracts/ConversionRates2.sol
@@ -205,7 +205,7 @@ contract ConversionRates2 is ConversionRates {
 
         if (len > 0) {
             fx = f.x[len - 1];
-            if (fx > fx) {
+            if (fx < 0) {
                 change += fx * f.y[len - 1];
             }
         }
--- a/contracts/ConversionRates2.sol
+++ b/contracts/ConversionRates2.sol
@@ -1,125 +1,15 @@
 pragma solidity 0.4.18;
 
-
 import "./ERC20Interface.sol";
-import "./VolumeImbalanceRecorder.sol";
-import "./Utils.sol";
-import "./ConversionRatesInterface.sol";
+import "./ConversionRates.sol";
 
 /// @title ConversionRates2 contract - new ConversionRates contract with step function enhancement
 /// Also fixed issue: https://github.com/KyberNetwork/smart-contracts/issues/291
 
-contract ConversionRates2 is ConversionRatesInterface, VolumeImbalanceRecorder, Utils {
+contract ConversionRates2 is ConversionRates {
 
-    // bps - basic rate steps. one step is 1 / 10000 of the rate.
-    struct StepFunction {
-        int[] x; // quantity for each step. Quantity of each step includes previous steps.
-        int[] y; // rate change per quantity step  in bps.
-    }
-
-    struct TokenData {
-        bool listed;  // was added to reserve
-        bool enabled; // whether trade is enabled
-
-        // position in the compact data
-        uint compactDataArrayIndex;
-        uint compactDataFieldIndex;
-
-        // rate data. base and changes according to quantity and reserve balance.
-        // generally speaking. Sell rate is 1 / buy rate i.e. the buy in the other direction.
-        uint baseBuyRate;  // in PRECISION units. see KyberConstants
-        uint baseSellRate; // PRECISION units. without (sell / buy) spread it is 1 / baseBuyRate
-        StepFunction buyRateQtyStepFunction; // in bps. higher quantity - bigger the rate.
-        StepFunction sellRateQtyStepFunction;// in bps. higher the qua
-        StepFunction buyRateImbalanceStepFunction; // in BPS. higher reserve imbalance - bigger the rate.
-        StepFunction sellRateImbalanceStepFunction;
-    }
-
-    /*
-    this is the data for tokenRatesCompactData
-    but solidity compiler optimizer is sub-optimal, and cannot write this structure in a single storage write
-    so we represent it as bytes32 and do the byte tricks ourselves.
-    struct TokenRatesCompactData {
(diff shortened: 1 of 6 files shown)


contracts/ConversionRates2.sol
```diff
@@ -0,0 +1,426 @@
...
+            if (x > lastStepAmount) {
+                change += (x - lastStepAmount) * f.y[len - 1];
+            }
+            return change / x;
+        }
+        // revert steps order for negative number
+        int lastStepBps = 0;
+        uint id;
+        for (ind = len; ind > 0; ind--) {
+            id = ind - 1;
```
ilanDoron: not sure why you need both ind and id
seems like ind isn't used at all in the loop

contracts/ConversionRates2.sol
```diff
@@ -0,0 +1,426 @@
...
+                change += (f.x[ind] - lastStepAmount) * f.y[ind];
+                lastStepAmount = f.x[ind];
+            }
+            if (x > lastStepAmount) {
+                change += (x - lastStepAmount) * f.y[len - 1];
+            }
+            return change / x;
+        }
+        // revert steps order for negative number
+        int lastStepBps = 0;
```
ilanDoron: better to avoid assuming last step bps is 0 for first iteration
it could have some value.

contracts/ConversionRates2.sol
```diff
@@ -0,0 +1,426 @@
...
+        require(xBuy.length <= MAX_STEPS_IN_FUNCTION);
+        require(xSell.length <= MAX_STEPS_IN_FUNCTION);
+        require(tokenData[token].listed);
+
+        if (xBuy.length > 0) {
+            // verify qties are increasing
+            require(xBuy[0] >= 0);
+            require(xSell[0] >= 0);
+            for(uint i = 0; i < xBuy.length - 1; i++) {
+                require(xBuy[i] < xBuy[i + 1]);
```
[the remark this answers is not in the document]
manhlx3006: Updated

contracts/ConversionRates2.sol
```diff
@@ -0,0 +1,426 @@
...
+    {
+        require(xBuy.length == yBuy.length);
+        require(xSell.length == ySell.length);
+        require(xBuy.length <= MAX_STEPS_IN_FUNCTION);
+        require(xSell.length <= MAX_STEPS_IN_FUNCTION);
+        require(tokenData[token].listed);
+
+        if (xBuy.length > 1) {
+            // verify qties are increasing
+            for(uint i = 0; i < xBuy.length - 1; i++) {
```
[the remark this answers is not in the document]
manhlx3006: Updated

contracts/ConversionRates2.sol
manhlx3006: yeah alr aware of it, for example qty is -10, steps are
[10, 20, 30]
[50, 100, 500]
what should be the result? 0 or 50?
similarly if steps are
[-100, -20, 10, 30]
[-50, -10, 20, 50]
and qty is -10, what is the result of bps?

ilanDoron: result in those cases should be the same results we would have got with old logic.

test/conversionRate2.js
```diff
@@ -0,0 +1,1410 @@
...
+let convRatesInst;
+
+contract('ConversionRates2', function(accounts) {
+    it("should init globals", function() {
+        admin = accounts[0];
+        alerter = accounts[1];
+        operator = accounts[2];
+        reserveAddress = accounts[3];
+    })
+    it("should test bytes14.", async function () {
```
ilanDoron: Can remove this test.
tested in conversionRate.

contracts/ConversionRates2.sol
```diff
@@ -0,0 +1,215 @@
...
+
+    function executeStepFunction(StepFunction f, int x) internal pure returns(int) {
+        uint len = f.y.length;
+        uint ind;
+        if (x == 0) {
+            // fallback to old logics
+            for(ind = 0; ind < len; ind++) {
+                if (x <= f.x[ind]) { return f.y[ind]; }
+            }
+            return f.x[len - 1];
```
manhlx3006: should we add for `len` check here? or just check if `len == 0` after defined `uint len = f.y.length`

test/conversionRate2.js
```diff
@@ -0,0 +1,1456 @@
...
+        assert.equal(basicData[1].valueOf(), 0, "token should not be enabled");
+
+        //enable token and see enabled
+        await convRatesInst.setTokenControlInfo(token.address, minimalRecordResolution, maxPerBlockImbalance, maxTotalImbalance);
+        await convRatesInst.enableTokenTrade(token.address);
+        basicData = await convRatesInst.getTokenBasicData(token.address);
+        assert.equal(basicData[1].valueOf(), 1, "token should be enabled");
+    });
+
+    it("should get buy rate with update according to compact data update.", async function () {
```
ilanDoron: can remove all tests not relevant to code in conversionRate2

test/conversionRate2.js
```diff
@@ -0,0 +1,1467 @@
...
+        assert.equal(blockNum.valueOf(), currentBlock.valueOf(), "bad block number returned");
+
+        blockNum = await convRatesInst.getRateUpdateBlock(tokens[11]);
+
+        assert.equal(blockNum.valueOf(), currentBlock.valueOf(), "bad block number returned");
+    });
+
+
+    it("should verify setCompactData reverted when block number out of range.", async function () {
+        let block = 0xffffffff1; //block number limited to size of int
```
ilanDoron: don't need this test here.
leave only tests relevant to code of this contract...

Issue #425, opened 2019-08-26
Check knc arbitrage for setKncRate in fee burner should be moved to feeBurner contract. Currently checked in getExpectedRate


ilanDoron:
Currently set KNC rate in fee Burner can be called by anyone.
some of the logic includes avoid setting rate when arbitrage exists.
this test was added in expected rate contract. to avoid deploying fee burner again. When fee burner is deployed, this test should be removed from expected rate and added to fee Burner.

ilanDoron:
This is the code in getExpectedRate
its a workaround that should be replaced.


if (src == knc &&
            dest == ETH_TOKEN_ADDRESS &&
            srcQty == UNIT_QTY_FOR_FEE_BURNER )
        {
            if (checkKncArbitrageRate(expectedRate)) expectedRate = 0;
        }


and


function checkKncArbitrageRate(uint currentKncToEthRate) public view returns

ilanDoron:
in Katalyst, fee burner contract is removed.
so not relevant.

PR #433, opened 2019-09-06
Arbitrage check shifted to FeeBurner

Anyhowclick:
[Fix for issue 425](https://github.com/KyberNetwork/smart-contracts/issues/425)
- L90 of ExpectedRate.sol translated to L94 of FeeBurner.sol
- L92 of ExpectedRate.sol translated to L95 of FeeBurner.sol
- L93 of ExpectedRate.sol translated to L98 of FeeBurner.sol 
- Old line `require(kyberEthKncRate * kyberKncEthRate < PRECISION ** 2 * 2);` is a looser check of the new check, so I removed it.
- Removed old line 97 of FeeBurner.sol, as it's covered under L95.

- The tests broke because `setKncRate()` now fails instead of the expectedRate contract returning zero, so they had to be fixed. 
- The last test in `FeeBurner.js`, `should verify function 'setKncRate' will fail if last call to findBestRate fails.`, is passing where < 2.8M gas is needed.... Not sure why this is the case, or if this is to be expected.


contracts/FeeBurner.sol
```diff
@@ -90,11 +90,15 @@ contract FeeBurner is Withdrawable, FeeBurnerInterface, Utils2 {
         (kyberEthKncRate, ) = kyberNetwork.getExpectedRate(ETH_TOKEN_ADDRESS, ERC20(knc), (10 ** 18));
         (kyberKncEthRate, ) = kyberNetwork.getExpectedRate(ERC20(knc), ETH_TOKEN_ADDRESS, (10 ** 18));
 
+        // if returned rate is 0, can't trust this arb test. assume we have arb
+        require(kyberEthKncRate > 0);
```
ilanDoron: seems redundant to line 100
no?

contracts/FeeBurner.sol
```diff
@@ -90,11 +90,15 @@ contract FeeBurner is Withdrawable, FeeBurnerInterface, Utils2 {
...
+        require(kyberEthKncRate <= MAX_RATE && kyberKncEthRate <= MAX_RATE);
+
+        //check for arbitrage
```
ilanDoron: its actually internal arbitrage. right?

test/feeBurner.js
```diff
@@ -644,9 +640,96 @@ contract('FeeBurner', function(accounts) {
...
+
+        let burnerInitialkncRate = await feeBurnerInst.kncPerEthRatePrecision();
+//         has arb. get rate will return 0. should revert.
+        try {
+            let rxx = await feeBurnerInst.setKNCRate();
+            assert(false, "throw was expected in line above.")
+        } catch(e){
+            assert(Helper.isRevertErrorMessage(e), "expected throw but got: " + e);
+        }
+
```
ilanDoron: can please fix rate here and see set rate does succeed

test/feeBurner.js
```diff
@@ -644,9 +640,96 @@ contract('FeeBurner', function(accounts) {
...
+        const kncToEthRatePrecision = precisionUnits.div(aRate);
 
+        await mockKyberNetwork.setPairRate(ethAddress, kncToken.address, ethToKncRatePrecision);
+        await mockKyberNetwork.setPairRate(kncToken.address, ethAddress, kncToEthRatePrecision);
+
+        let rxEthToKnc = await mockKyberNetwork.getExpectedRate(ethAddress, kncToken.address, 10 ** 19);
+        let rxKncToEth = await mockKyberNetwork.getExpectedRate(kncToken.address, ethAddress, 10 ** 19);
+        assert.equal(rxEthToKnc[0].valueOf(), ethToKncRatePrecision.floor().valueOf())
+        assert.equal(rxKncToEth[0].valueOf(), kncToEthRatePrecision.floor().valueOf())
+
```
ilanDoron: can test here failing rate before setting non failing rate (no arb)

test/feeBurner.js
```diff
@@ -644,9 +640,96 @@ contract('FeeBurner', function(accounts) {
...
+        // no arb. set rate should not fail
+        let rx = await feeBurnerInst.setKNCRate();
+        burnerkncRateAfterSet = await feeBurnerInst.kncPerEthRatePrecision();
+        assert.equal(ethToKncRatePrecision2.valueOf(), burnerkncRateAfterSet.valueOf());
+    });
+
+    it("should verify function 'setKncRate' will fail if last call to findBestRate fails.", async function() {
+        mockKyberNetwork = await NetworkFailingGetRate.new();
```
ilanDoron: so as discussed. this test can be removed.

contracts/FeeBurner.sol
Anyhowclick: Ah, because it's covered in line 100. Agreed. Can remove then.

contracts/FeeBurner.sol
Anyhowclick: Yes. I'll rewrite the comment there.

test/feeBurner.js
Anyhowclick: =)

test/feeBurner.js
Anyhowclick: Ah, I remember now. This used to be 1 test together with the next test, but I separated them because of the network timeout issue I had with Buidler. 

I can try to combine the test again with the next and run it, but I expect I'll run into the same issue.

test/feeBurner.js
Anyhowclick: As mentioned above.

test/feeBurner.js
ilanDoron: Is the test working now?

ilanDoron: Anyway.
don't move back to previous setup
if its tested else where its fine.

test/feeBurner.js
ilanDoron: OK
lets keep it.

test/feeBurner.js
Anyhowclick: Ah oops, didn't see this. Yeah they're tested.

Issue #452, opened 2019-10-24
remove multisig support from wrapConversionRate


ilanDoron:
Currently wrap conversion rate has multisig support.
will remove this so this contract only used for setting data
it will have one admin which will be a multisig contract
so all data settings are done with one admin

ilanDoron:
merged to master
passed staging.
will be used in next production updates

Issue #441, opened 2019-09-18
re organize reserve smart contracts in folders


ilanDoron:
reserve folder structure
high level: reserve folder
reserveBridge folder. under this have oasis, uniswap, dutchx

fprReserve

aprReserve

orderBookReserve
inside orderbook we will have permissionless folder

ilanDoron:
text edited.
re opening

ilanDoron:
there was one issue that delayed this.
how to verify on line deployed contracts.
if we change the text in files, the byte code result changes - since it includes not only byte code but some text reference.

solution for this:
run sha3 on all contracts with current folder structure.
when verifying contracts. get byte code. run sha3 and compare it to saved sha3 values.

PR #453, opened 2019-10-24
Organize folder structure and update WrapConversionRate.sol #441 && #452

ilanDoron:
this includes organizing audit folder and organizing reserves.
Remove multisig support from WrapConversionRate.sol according to issue #452


contracts/kyberReserves/aprConversionRate/LiquidityConversionRates.sol
```diff
@@ -1,9 +1,10 @@
 pragma solidity 0.4.18;
 
-import "./ConversionRatesInterface.sol";
+
```
Anyhowclick: Extra line here (linter error fix I think)

contracts/wrapperContracts/WrapConversionRate.sol
```diff
@@ -2,246 +2,89 @@ pragma solidity 0.4.18;
...
-    uint constant internal TOKEN_INFO_DATA_INDEX = 1;
-    uint constant internal VALID_DURATION_DATA_INDEX = 2;
-    uint constant internal NUM_DATA_INDEX = 3;
-
     //general functions
     function WrapConversionRate(ConversionRates _conversionRates, address _admin) public
-        WrapperBase(PermissionGroups(address(_conversionRates)), _admin, NUM_DATA_INDEX)
+        WrapperBase(PermissionGroups(address(_conversionRates)), _admin, 0)
     {
-        require(_conversionRates != address(0));
```
Anyhowclick: This check isn't needed anymore?

contracts/wrapperContracts/WrapConversionRate.sol
```diff
@@ -2,246 +2,89 @@ pragma solidity 0.4.18;
...
-        uint nonce;
-        (signatures, nonce) = getDataTrackingParameters(VALID_DURATION_DATA_INDEX);
-        return(signatures);
-    }
-
-    function performAddToken() internal {
-        conversionRates.addToken(addTokenData.token);
+        conversionRates.addToken(token);
 
         conversionRates.addOperator(this);
```
Anyhowclick: Why can the `addOperator` be called multiple times for one address. Isn't there a require check to prevent duplicates, no? 
`require(!operators[newOperator]); // prevent duplicates.`

Anyhowclick: I see that the function adds this contract as an operator, then removes it. It's for security reasons?

contracts/wrapperContracts/WrapConversionRate.sol
```diff
@@ -2,246 +2,89 @@ pragma solidity 0.4.18;
...
+                 maxPerBlockImbalanceValues[i],
+                maxTotalImbalanceValues[i]);
         }
     }
-}
 
+    //valid duration blocks
+    ///////////////////////
+    function setValidDurationData(uint validDurationBlocks) public onlyAdmin {
+        require(validDurationBlocks > 5);
```
Anyhowclick: Would it be better to set the `5` as a public constant? Or variable if we consider ETH2.0 :p

contracts/wrapperContracts/WrapConversionRate.sol
ilanDoron: actually the remove isn't really necessary
was thinking not to have extra operator addresses if not needed...

contracts/wrapperContracts/WrapConversionRate.sol
ilanDoron: possible.
will set public constant.

its all very internal stuff.

PR #457, opened 2019-10-29
Istanbul Fix

Anyhowclick:


contracts/KyberNetwork.sol
```diff
@@ -67,10 +67,7 @@ contract KyberNetwork is Withdrawable, Utils2, KyberNetworkInterface, Reentrancy
     event EtherReceival(address indexed sender, uint amount);
 
     /* solhint-disable no-complex-fallback */
-    // To avoid users trying to swap tokens using default payable function. We added this short code
-    //  to verify Ethers will be received only from reserves if transferred without a specific function call.
     function() public payable {
-        require(reserveType[msg.sender] != ReserveType.NONE);
         EtherReceival(msg.sender, msg.value);
```
ilanDoron: for sure some tests should change here...
did u check.

I imagine we check only listed reserves can pay network.

Anyhowclick: Ermmm... I just ran both the KNP and KN test scripts. No such check done =O

ilanDoron: OK
I see we really have no tests fro this.

PR #461, opened 2019-10-30
Enable wrapConversionRate to control all adminOnly functions in convesion rate

ilanDoron:


contracts/wrapperContracts/WrapConversionRate.sol
```diff
@@ -56,6 +56,18 @@ contract WrapConversionRate is WrapperBase {
...
 
+    // enable trade per token
+    //////////////////////
+    function enableTokenTrade(ERC20 token) public onlyAdmin {
+        conversionRates.enableTokenTrade(token);
+    }
+
+    // set conversion rates reserve address
+    //////////////////////
+    function setReserveAddress(address reserve) public onlyAdmin {
```
manhlx3006: we alr had `setReserveAddress(address) public onlyAdmin` in `ConversionRate.sol`, why do we need this func here in wrapper? Similar to `enableTokenTrade`

ilanDoron: since we set wrapper as admin for ConversionRate.sol
and it is the only admin

We don't want to switch admin for those specific function calls.

PR #470, opened 2019-11-01
[WIP]according to APR audit.

ilanDoron:
limit some configuration parameters.
and overflow checks.


contracts/reserves/aprConversionRate/LiquidityFormula.sol
```diff
@@ -147,6 +150,9 @@ contract LiquidityFormula is UtilMath {
     {
         uint pe = pE(r, pMIn, e, precision);
         uint rpe = r * pe;
+
+        require(!checkMultOverflow(rpe, deltaT));
+        require(precision * precision + rpe * deltaT/precision > precision * precision);
```
manhlx3006: In line 163, we must check r to be non-zero before using `/`

manhlx3006: Similar to in func `countLeadingZeros`, there is `if (p/(q*denomator) > 0) return uint(i);` which also need to check `q` to be non-zero.

ilanDoron: r in FP is checked to be > 0 when set.
and later it is never modified.

ilanDoron: q is set to be precision * precision.
and precision parameter is set on and can't be zero (after my changes)

manhlx3006: but they are public call right? so we can call with 0 value unless it is internal func

ilanDoron:
this PR only for code compare.
APR will be changed only after final APR report

PR #472, opened 2019-11-05
handle zero qty in get rate permission less reserve

ilanDoron:


test/orderbookReserve.js
```diff
@@ -2571,8 +2571,7 @@ contract('OrderbookReserve', async (accounts) => {
             assert.equal(list[0].valueOf(), sellOrder1ID);
             assert.equal(list[1].valueOf(), sellOrder2ID);
 
-            let expectedGasDiff4BatchOrders = 100000;
-
+            let expectedGasDiff4BatchOrders = 65000;
```
manhlx3006: why do we have this change here?

Anyhowclick: Test doesn't fail with original value of 100000, so this change is unnecessary.

ilanDoron: Oh yes.
its leftover.
will remove.
thanks

Issue #434, opened 2019-09-11
avoid srcAmount 0 on token to token trades


ilanDoron:
When calling getConversionRate with srcAmount 0, some reserves revert.
On simple trade (not token to token) getExpectedRate modifies srcAmount 0 to 1, to avoid divide by 0.
Token to token trades are run in two steps.
step 1 - token1 to eth.
step 2 - eth to token2.
If srcAmount on Token to eth is small, the Eth result will be 0 when calling step 2 eth to token.
this will cause a revert.

ilanDoron:
Currently this revert is handled in get expected rate with assembly code that blocks tx from revert.
Can fix this issue between step 1 and step 2 of token to token. translate amount 0 to 1. or return 0 rate when Eth amount is 0.

after fix, remove expected rate assembly code...

PR #483, opened 2019-11-12
Zero Src Qty Handling

Anyhowclick:
Fix to issue #434

One issue I want to raise: For token -> token rate queries with 0 src qty, while `network.findBestRate` and `network.findBestRateOnlyPermission` returns zero, `network.getExpectedRate` **will revert** if the reserve fails to handle zero src qty properly, because of the `expectedRate.expectedRateSmallQty` function, which calls `network.searchBestRate` directly.

To be clear, trades will **not revert** because the trade input checks that `srcAmount != 0`, and it uses the `findBestRateTokenToToken` function (where the fix lies), which is the underlying function used by `findBestRate` and `findBestRateOnlyPermission`.

It boils down to how we want to deal with zero source quantity rate queries. Do we want to return something, or just return zero?


contracts/KyberNetwork.sol
```diff
@@ -436,13 +436,20 @@ contract KyberNetwork is Withdrawable, Utils3, KyberNetworkInterface, Reentrancy
             searchBestRate(src, ETH_TOKEN_ADDRESS, srcAmount, usePermissionless);
 
         result.weiAmount = calcDestAmountWithDecimals(srcDecimals, ETH_DECIMALS, srcAmount, result.rateSrcToEth);
+        //if weiAmount is zero, return zero rate to avoid revert in ETH -> token call
+        if (result.weiAmount == 0) {
+            result.reserve2 = address(0);
+            result.rateEthToDest = 0;
+            result.destAmount = 0;
+            result.rate = 0;
```
ilanDoron: please write this in a way that created minimum code changes.
exapmle: if weiAmount is zero.
try to set result and return it. and leave rest of code untouched.

contracts/ExpectedRate.sol
```diff
@@ -88,55 +105,4 @@ contract ExpectedRate is Withdrawable, ExpectedRateInterface, Utils2 {
...
-                0x40) // Output size are (uint, uint) = 64 bytes
-
-            rate := mload(add(x,0x84))  //Assign 2nd output to rate, first output not used,
-            mstore(0x40,x)    // Set empty storage pointer back to start position
-        }
-
-        if (success != 1) didRevert = true;
-    }
-}
+}
```
ilanDoron: CR LF on last line

contracts/ExpectedRate.sol
```diff
@@ -54,10 +54,27 @@ contract ExpectedRate is Withdrawable, ExpectedRateInterface, Utils2 {
...
         if (srcQty == 0) srcQty = 1;
 
-        bool didRevert = false;
+        //not needed
+        uint bestReserve;
 
-        (didRevert, expectedRate, slippageRate) = safeFindBestRate(src, dest, srcQty, usePermissionless);
-        if (didRevert) return (0, 0);
+        if (usePermissionless) {
+            (bestReserve, expectedRate) = kyberNetwork.findBestRate(src, dest, srcQty);
```
ilanDoron: best if can write the code as it was before safeFindBestRate was added.
is it?

test/expectedRate.js
```diff
@@ -522,6 +522,37 @@ contract('ExpectedRate', function(accounts) {
         assert.equal(rates[1].valueOf(), 0, "unexpected rate");
     });
 
+    it("should verify when src qty 0, network.findBestRate and network.findBestRateOnlyPermission don't revert", async function() {
+        let tokenSrcInd = 2;
+        let tokenDestInd = 1;
+        let qty = 0;
+
+        //create malicious reserve
+        let badReserve = await MaliciousReserve.new(network.address, pricing1.address, admin);
```
ilanDoron: don't get why reserve is called malicious?
does it attack network?
or only revert when queried with 0 quantity?

contracts/ExpectedRate.sol
Anyhowclick: https://github.com/KyberNetwork/smart-contracts/blob/37e9bcd67dd8799ea2f571c26ecc33a6945b95aa/contracts/ExpectedRate.sol

It's the same as this, right? Minus the check for knc arbitrage.

test/expectedRate.js
Anyhowclick: It reverts when queried with 0 quantity

contracts/KyberNetwork.sol
Anyhowclick: Updated!

contracts/ExpectedRate.sol
```diff
@@ -54,10 +54,27 @@ contract ExpectedRate is Withdrawable, ExpectedRateInterface, Utils2 {
...
+        } else {
+            (bestReserve, expectedRate) = kyberNetwork.findBestRateOnlyPermission(src, dest, srcQty);
+
+            if (quantityFactor != 1) {
+                (bestReserve, slippageRate) = kyberNetwork.findBestRateOnlyPermission(src, dest,
+                    (srcQty * quantityFactor));
+            } else {
+                slippageRate = expectedRate;
+            }
+        }
```
ilanDoron: regarding your PR comment.
I Agree expectedRateSmallQty could fail.
solution would be after this line:
uint ethQty = calcDestAmount(src, ETH_TOKEN_ADDRESS, srcQty, rateSrcToEth)

if ethQty == 0, can use ethqty = 1 or return 0.
both are fine from my side.

contracts/ExpectedRate.sol
ilanDoron: Great
thanks

contracts/ExpectedRate.sol
Anyhowclick: Added!

PR #484, opened 2019-11-12
Fix npm issues, remove unused packages, fix npm and node version

manhlx3006:


package.json
```diff
@@ -37,22 +36,10 @@
...
-    "prettier": "^1.15.3",
-    "prettier-eslint-cli": "^4.7.1",
-    "prettier-standard": "^8.0.1",
-    "solidity-coverage": "0.4.15",
-    "truffle-assertions": "^0.7.2",
-    "truffle-flattener": "^1.2.5"
+    "solidity-coverage": "^0.6.7"
+  },
+  "engines": {
+    "node": ">=12.13.0",
```
ilanDoron: I don't think this is the case before buidler.
all works for you?

PR #485, opened 2019-11-13
[455] Enable unlist orderbook permissionless reserve

manhlx3006:


contracts/reserves/orderBookReserve/permissionless/OrderbookReserve.sol
```diff
@@ -660,6 +660,15 @@ contract OrderbookReserve is OrderIdManager, Utils2, KyberReserveInterface, Orde
         return getList(list);
     }
 
+    // return true if both sell and buy orders are empty
+    function isOrderbookEmpty() public view returns(bool) {
```
manhlx3006: If I remove `(, isEmpty) = tokenToEthList.getFirstOrder();` or `(, isEmpty) = ethToTokenList.getFirstOrder();`, then it works as normal
If I keep both, all tests in `permissionlessOrderbookReserveLister` are failed (init `PermissionlessOrderbookReserveLister` contract is out of gas)

PR #486, opened 2019-11-14
Added deployer script for minor network upgrade

Anyhowclick:


web3deployment/deployerv3.js
```diff
@@ -0,0 +1,473 @@
...
+  const output = await solc.compile({ sources: input }, 1);
+  // console.log(output);
+  console.log("finished compilation");
+
+  if (!dontSendTx) {
+    await waitForEth();
+  }
+
+  console.log("deploying kyber network");
+  [networkAddress,networkContract] = await deployContract(output, "KyberNetwork.sol:KyberNetwork", [sender]);
```
ilanDoron: after each deploy please print address as output.
since if deploy script fails in middle of deploy, its easier to restart a deploy and reuse some of the deployed contracts.

web3deployment/deployerv3.js
```diff
@@ -0,0 +1,473 @@
...
+  await sendTx(feeBurnerContract.methods.setKNCRate());
+  console.log("set tax fees bps");
+  await sendTx(feeBurnerContract.methods.setTaxInBps(taxFeesBps));
+  if(taxWalletAddress != '' && taxWalletAddress != 0) {
+    console.log("set tax wallet address");
+    await sendTx(feeBurnerContract.methods.setTaxWallet(taxWalletAddress));
+  }
+
+  //set wallet fees for non-standard fees
+  for(let i = 0 ; i < walletDataArray.length; i++) {
```
ilanDoron: great.
did u try reading events?
did u find some of those wallets with fee > 30

web3deployment/deployerv3.js
```diff
@@ -0,0 +1,473 @@
...
+
+  myContract.options.address = address;
+
+  return [address,myContract];
+}
+
+const contractPath = path.join(__dirname, "../contracts/");
+
+const input = {
+  "ConversionRatesInterface.sol" : fs.readFileSync(contractPath + 'ConversionRatesInterface.sol', 'utf8'),
```
ilanDoron: please create a separate file that exports this input data.

I found myself fixing it so many times.

So it can be reused many times and fixed in one central place.

or if we have any other idea to use this list in all deploy scripts with min work

web3deployment/deployerv3.js
Anyhowclick: Ah, so u want it after every deploy, ok.

web3deployment/deployerv3.js
Anyhowclick: Yup, I did. I didn't add that script... should I?

web3deployment/deployerv3.js
Anyhowclick: Ahhhh.... that's true. Ok, good idea to put it in a separate script.

Anyhowclick: I'm not sure how that would work if we have both v4 and v5 versions though... Hmm

ilanDoron: well
at least have the v4 inputs in a separate input file.
We will have to worry about v5 later.

PR #488, opened 2019-11-14
Revert "Arbitrage check shifted to FeeBurner"

ilanDoron:
Reverts KyberNetwork/smart-contracts#433

was decided not to include this in next deploy
will be delayed.

### DIFF
--- a/test/kyberNetwork.js
+++ b/test/kyberNetwork.js
@@ -2809,13 +2809,14 @@ contract('KyberNetwork', function(accounts) {
             let srcAmountTwei = 136;
             let maxDestAmount = (new BigNumber(10)).pow(18);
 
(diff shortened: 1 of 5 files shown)


test/kyberNetwork.js
```diff
@@ -2809,14 +2809,15 @@ contract('KyberNetwork', function(accounts) {
             let srcAmountTwei = 136;
             let maxDestAmount = (new BigNumber(10)).pow(18);
 
-            buyRate = await network.getExpectedRate(tokenSrc.address, tokenDest.address, srcAmountTwei.valueOf());
```
Anyhowclick: The changes here clashes with the same src and dest PR... Was working on multiple features on the same branch last time, so I think these changes were here by accident. Could u leave this file untouched? =)

PR #506, opened 2019-11-25
Ilan/extra fpr prints

ilanDoron:
PR for reference
don't pull


ilanDoron:
this was not for pulling

PR #508, opened 2019-11-27
fixed FPR deployment script

STYJ:


web3deployment/reserveDeployer.js
```diff
@@ -98,26 +98,109 @@ async function deployContract(solcOutput, contractName, ctorArgs) {
...
-  "KyberNetwork.sol" : fs.readFileSync(contractPath + 'KyberNetwork.sol', 'utf8'),
-  "WhiteList.sol" : fs.readFileSync(contractPath + 'WhiteList.sol', 'utf8'),
-  "KyberReserveInterface.sol" : fs.readFileSync(contractPath + 'KyberReserveInterface.sol', 'utf8'),
-  "Withdrawable.sol" : fs.readFileSync(contractPath + 'Withdrawable.sol', 'utf8'),
-  "KyberReserve.sol" : fs.readFileSync(contractPath + 'KyberReserve.sol', 'utf8'),
-  "Wrapper.sol" : fs.readFileSync(contractPath + 'mockContracts/Wrapper.sol', 'utf8')
-};
+  'ConversionRatesInterface.sol': fs.readFileSync(
+    contractPath + 'ConversionRatesInterface.sol',
+    'utf8'
```
ilanDoron: please use the compile contracts file from branch development.
Can add it to this deploy.
if its lacking any contract please add it.

web3deployment/reserveDeployer.js
```diff
@@ -153,9 +236,9 @@ function parseInput( jsonInput ) {
 
       tokens.push(address);
       const dict = {
-        minimalRecordResolution : web3.utils.toBN(val["minimalRecordResolution"]),
-        maxPerBlockImbalance : web3.utils.toBN(val["maxPerBlockImbalance"]),
-        maxTotalImbalance : web3.utils.toBN(val["maxTotalImbalance"])
+        minimalRecordResolution : val["minimalRecordResolution"],
+        maxPerBlockImbalance : val["maxPerBlockImbalance"],
+        maxTotalImbalance : val["maxTotalImbalance"]
       };
```
ilanDoron: I don't follow, why don't we need BN here anymore?
or it was never needed?

STYJ: The web3.utils.toBn function returns a BN object. Passing this into the function will also fail and return an error like `<rejected> Error: invalid number value (arg="minimalRecordResolution", coderType="uint256", value="38d7ea4c68000")`

The values stored in the config file (ropsten.json) is already a string, there is no need to convert it to BN.

web3deployment/reserveDeployer.js
STYJ: are we referring to this? https://github.com/KyberNetwork/smart-contracts/blob/development/compilation.sh

web3deployment/reserveDeployer.js
STYJ: As for the package-lock, I wasn't sure why it changed when I did npm-install. Anyway, I will remove it.

web3deployment/reserveDeployer.js
ilanDoron: this one:
https://github.com/KyberNetwork/smart-contracts/blob/development/web3deployment/compileContracts.js

web3deployment/reserveDeployer.js
ilanDoron: yes please remove it.
we have started using npm ci for a new repo clone
and not npm i.

anyway package lock shouldn't be added to PRs unless packages changed as well.

STYJ: Understood. Thanks for letting me know!

PR #518, opened 2019-12-05
fix uniswap bridge for small amounts src qty (1)

ilanDoron:


contracts/reserves/bridgeReserves/uniswap/KyberUniswapReserve.sol
```diff
@@ -558,6 +558,9 @@ contract KyberUniswapReserve is KyberReserveInterface, Withdrawable, Utils2 {
         UniswapExchange exchange;
         if (src == ETH_TOKEN_ADDRESS) {
             exchange = UniswapExchange(tokenExchange[address(dest)]);
+
+            if (deductFee(srcQty) == 0) return (0, 0);
```
manhlx3006: should use a local variable for deductFee(srcQty) so we don't have to call twice

PR #520, opened 2019-12-09
Clicky/buidl integration

ilanDoron:


ilanDoron:
will be done on another PR

PR #531, opened 2019-12-19
Clicky's web3 compatibility upgrade

Anyhowclick:


manhlx3006:
In general you should remove all `.valueOf()` from big number when using new bn.js lib as it doesn't have this func anymore.

test/v4/kyberUniswapReserve.js
```diff
@@ -715,12 +711,12 @@ contract("KyberUniswapReserve", async accounts => {
                 1 /* eth */,
                 1 /* token */
             );
-            const amount = web3.toWei("1");
-            const conversionRate = new BigNumber(10).pow(18);
+            const amount = web3.utils.toWei("1");
+            const conversionRate = new BN(10).pow(new BN(18));
```
ilanDoron: for next files.
can do
new BN(10 ** 18)
also other combinations can  be shorter like this.

PR #537, opened 2019-12-24
wrapper files(3), orderIdManager, orderList, mockDepositxx, mockExchnagexxx, permission less order book reserve lister

ilanDoron:


test/v4/orderList.js
```diff
@@ -1821,3 +1865,7 @@ async function debugOrders(max) {
             );
     }
 }
+
+function log(str) {
```
manhlx3006: this func is not used again

ilanDoron: well
I am thinking anytime I start debugging I will use it.
so its easier for me it just stays there....

Think its not too ugly...

PR #550, opened 2020-01-03
Add findRatesAndAmounts

Anyhowclick:


contractsV5/KyberNetwork.sol
```diff
@@ -508,7 +508,8 @@ contract KyberNetwork is Withdrawable, Utils, IKyberNetwork, ReentrancyGuard {
         uint totalFeeWei;
         uint tradeWeiAmount;
         uint destAmount;
-        uint rate; //the accumulated rate for full trade
+        uint rateWithFee; // rate for full trade, after accounting for fees
```
ilanDoron: Seems to me we can have:
destAmountWithFee
destAmountNoFee

or 
like here.
no need for both.

In general the code now works with amounts.
if we move to rates. need to check it works nice in other parts of the code

contractsV5/KyberNetwork.sol
```diff
@@ -508,7 +508,8 @@ contract KyberNetwork is Withdrawable, Utils, IKyberNetwork, ReentrancyGuard {
         uint totalFeeWei;
         uint tradeWeiAmount;
```
ilanDoron: seems here in wei amount should also define if withFee
or 
noFee

contractsV5/KyberNetwork.sol
```diff
@@ -523,28 +524,107 @@ contract KyberNetwork is Withdrawable, Utils, IKyberNetwork, ReentrancyGuard {
...
+        }
 
-        // else find best rate
+        //add percentage for ETH -> token reserves
+        //if there is no ETH -> token reserve specified, percentage added will be zero, to be handled after searching best rate
+        for (i = 0; i < tradeData.ethToToken.addresses.length; i++) {
+            if (tradeData.ethToToken.isPayingFees[i]) {
+                tradeData.feePayingPercentage += (i == tradeData.ethToToken.splitValuesPercent.length-1) ? (100 - percentageSoFar) : tradeData.ethToToken.splitValuesPercent[i];
+            }
+            percentageSoFar += tradeData.tokenToEth.splitValuesPercent[i];
```
ilanDoron: why do we need the local one?
can't save all in tradeData.feePayingPercentage?

contractsV5/KyberNetwork.sol
```diff
@@ -523,28 +524,107 @@ contract KyberNetwork is Withdrawable, Utils, IKyberNetwork, ReentrancyGuard {
...
 
-        // else find best rate
-        // calculate accumulated amounts
-
-
-        // dest amount sbutract total fee wei
-
-        // calc final rate
+            //calculate destAmounts (with and without fees)
+            tradeData.destAmount = calcDestAmountWithDecimals(ETH_DECIMALS, tradeData.ethToToken.decimals, tradeWeiAmountAfterFee, rate);
```
ilanDoron: can we calculate this with
(tradeWeiAmount - weiFeeAmount)
and avoid local parameter?

contractsV5/KyberNetwork.sol
Anyhowclick: Ok. It's noFee.

contractsV5/KyberNetwork.sol
Anyhowclick: Hmmmm you're right, it's not needed.

contractsV5/KyberNetwork.sol
Anyhowclick: I assume you're referring to `tradeWeiAmountAfterFee`. We can, but `tradeWeiAmountAfterFee` is used quite a number of times. 6, to be exact.

contractsV5/KyberNetwork.sol
Anyhowclick: No it can't. Because we use N-1 instead of N for splitValuesPercent, I don't know the trade percentage of the last reserve without the variable. If I just store it in tradeData, then the feePayingPercentage will **always** be 100% if the last reserve is fee paying.

Example: 1st reserve is feeless (60% of tradeQty), 2nd reserve is fee paying (40% of tradeQty). 
Expected: feePayingPercentage = 40%
Actual:  feePayingPercentage = 100%

Anyhowclick: This can be solved if we store splitValuesPercent to be N... I think.

PR #554, opened 2020-01-09
Update findRatesAndAmounts and searchBestDestAmount back to searchBestRate

Anyhowclick:


contractsV5/KyberNetwork.sol
```diff
@@ -422,62 +423,61 @@ contract KyberNetwork is Withdrawable, Utils, IKyberNetwork, ReentrancyGuard {
...
+                dest,
+                (src == ETH_TOKEN_ADDRESS && isPayingFees) ? srcAmount * (BPS - takerFeeBps) / BPS : srcAmount, //srcAmountWithFee
                 block.number);
 
-            destAmounts[i] = srcAmountWithFee * rate / PRECISION;
-            destAmounts[i] = (dest == ETH_TOKEN_ADDRESS && isFeeLessReserve[reserveArr[i]])? destAmounts[i] : 
-                destAmounts[i] * (BPS - takerFeeBps) / BPS;
-            
-            if (destAmounts[i] > bestDestAmount) {
+            fullDestAmount = srcAmount * rates[i] / PRECISION;
```
ilanDoron: must user getDestAmount with decimals)

its my fault. I did it. :(
anyway better if u fix it in your PR

contractsV5/KyberNetwork.sol
```diff
@@ -422,62 +423,61 @@ contract KyberNetwork is Withdrawable, Utils, IKyberNetwork, ReentrancyGuard {
...
-        for (i = 0; i < reserveArr.length; i++) {
-            
+        for (uint i = 0; i < reserveArr.length; i++) {
+
             if (i == bestReserve) continue;
-            
-            if (destAmounts[i] > smallestRelevantDestAmount) 
-            {
+
+            fullDestAmount = srcAmount * rates[i] / PRECISION;
```
ilanDoron: must use calcDestQty
again my fault its not here.

ilanDoron: actually maybe its valid not to use it if we are consistent?
but if it is should be clarified some how.

contractsV5/KyberNetwork.sol
```diff
@@ -489,9 +489,7 @@ contract KyberNetwork is Withdrawable, Utils, IKyberNetwork, ReentrancyGuard {
             bestReserve = reserveCandidates[0];
         }
 
-        bestDestAmount = destAmounts[bestReserve];
-    
-        return (reserveArr[bestReserve], bestDestAmount, isFeeLessReserve[reserveArr[bestReserve]]);
+        return (reserveArr[bestReserve], rates[bestReserve], isFeePayingReserve[address(reserveArr[bestReserve])]);
```
ilanDoron: so what is the logic
comparing dest amount but return exact rate.
sound good.

so all amount modifications are elsewhere.

contractsV5/KyberNetwork.sol
```diff
@@ -546,17 +544,139 @@ contract KyberNetwork is Withdrawable, Utils, IKyberNetwork, ReentrancyGuard {
...
+                tradeData.feePayingReservesBps += BPS; //max percentage amount for ETH -> token
+                tradeData.numFeePayingReserves ++;
+            }
+
+            //calculate destAmountWithNetworkFee and actualDestAmount
+            tradeData.destAmountWithNetworkFee = calcDstQty(tradeData.tradeWei - tradeData.networkFeeWei, ETH_DECIMALS, tradeData.ethToToken.decimals, tradeData.ethToToken.rates[0]);
+            tradeData.actualDestAmount = calcDstQty(tradeData.tradeWei - tradeData.networkFeeWei - tradeData.platformFeeWei, ETH_DECIMALS, tradeData.ethToToken.decimals, tradeData.ethToToken.rates[0]);
+        }
+        // calc final rate
+        tradeData.rateWithNetworkFee = calcRateFromQty(srcAmount, tradeData.destAmountWithNetworkFee, tradeData.tokenToEth.decimals, tradeData.ethToToken.decimals);
```
ilanDoron: function is too long
solhin will surely complain.

can we have one function to serve the main loop in a generic way?

anyway.
don't have to fix for now.

contractsV5/KyberNetwork.sol
Anyhowclick: Errmmmm it means I will need the decimals.....

contractsV5/KyberNetwork.sol
Anyhowclick: My opinion is that it's valid not to, since we are only using the dest amount for comparison. Running issues with no. of local variables also....

contractsV5/KyberNetwork.sol
Anyhowclick: yes, that's correct.

contractsV5/KyberNetwork.sol
Anyhowclick: ok, TODO: refactor a bit hahaha

Anyhowclick: Done!

PR #555, opened 2020-01-11
orderbookReserve.js update to web3 v1.2.4

ayobuenavista:


test/v4/orderbookReserve.js
```diff
@@ -479,62 +476,61 @@ contract('OrderbookReserve', async (accounts) => {
...
             } catch(e) {
                 assert(Helper.isRevertErrorMessage(e), "expected throw but got: " + e);
             }
         })
 
         it("verify get update order hint reverts if Eth value is below min order value", async() => {
-            let tokenWeiDepositAmount = 30 * 10 ** 18;
-            let kncTweiDepositAmount = 600 * 10 ** 18;
-            let ethWeiDepositAmount = 3 * 10 ** 18;
+            let tokenWeiDepositAmount = PRECISION.mul(new BN(30));
```
ilanDoron: this isn't precision.

its confusing since our precision is same value as token decimals for many tokens.
so this should be one tokne.mul.(x)

test/v4/orderbookReserve.js
```diff
@@ -3268,501 +3187,491 @@ contract('OrderbookReserve', async (accounts) => {
...
+            Helper.assertEqual(orderList.length, 4);
         })
     });
 
     describe("knc stakes and burn", function() {
         it("maker add sell token order. see funds & knc stakes updated.", async () => {
-            let kncTweiDepositAmount = new BigNumber(600 * 10 ** 18);
-            let tokenWeiDepositAmount = new BigNumber(11.1 * 10 ** 18);
+            let kncTweiDepositAmount = PRECISION.mul(new BN(600));
+            let tokenWeiDepositAmount = new BN(String(PRECISION * 11.1));
```
ilanDoron: does string(PRECISION * 11.1) give same result as 11.1 * 10 ** 18 ??

ilanDoron: Oh
I see it is
nice solution.

test/v4/orderbookReserve.js
```diff
@@ -412,65 +409,65 @@ contract('OrderbookReserve', async (accounts) => {
         });
 
         it("verify get rate works only for token->Eth or eth->token. other options revert", async() => {
-            let rate = await reserve.getConversionRate(tokenAdd, ethAddress, 10 ** 18, 0);
-            assert.equal(rate.valueOf(), 0);
-            rate = await reserve.getConversionRate(ethAddress, tokenAdd, 10 ** 18, 0);
-            assert.equal(rate.valueOf(), 0);
+            let rate = await reserve.getConversionRate(tokenAdd, ethAddress, PRECISION, 0);
```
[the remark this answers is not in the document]
ayobuenavista: Sure, makes sense. How about "TWei" for token wei, since Wei is defined as 10 ** 18?

test/v4/orderbookReserve.js
ayobuenavista: Yeah, this is the simplest solution to make it work. BN doesn't understand decimals. So if you do `new BN(11.1)`, it will round it off to 11.

PR #561, opened 2020-01-15
Basic Trade Flow Functionality, remove whitelist 

Anyhowclick:


contractsV5/KyberHintParser.sol
```diff
@@ -0,0 +1,146 @@
...
+    bytes public constant SEPARATOR = "\x00";
+    bytes public constant MASK_IN_OPCODE = "\x01";
+    bytes public constant MASK_OUT_OPCODE = "\x02";
+    bytes public constant SPLIT_TRADE_OPCODE = "\x03";
+    bytes public constant END_OPCODE = "\xaa"; //consider using hint length instead
+
+    using BytesLib for bytes;
+    
+    struct ReservesHint {
+        bool isSplit; // to change to boolean flag to specify
```
ilanDoron: change to hint type
right?

contractsV5/KyberHintParser.sol
```diff
@@ -0,0 +1,146 @@
...
+        address[] addresses;
+        uint[] splitValuesBps;
+    }
+    
+    struct TradeHint {
+        ReservesHint ethToTokenReserves;
```
ilanDoron: do we need this?
or can set directly to trade data?

contractsV5/KyberNetwork.sol
```diff
@@ -22,10 +22,7 @@ interface IExpectedRate {
 ////////////////////////////////////////////////////////////////////////////////////////////////////////
 /// @title Kyber Network main contract
 contract KyberNetwork is Withdrawable, Utils, IKyberNetwork, ReentrancyGuard {
-
-    bytes public constant PERM_HINT = "PERM";
-    uint  public constant PERM_HINT_GET_RATE = 1 << 255; // for get rate. bit mask hint.
```
ilanDoron: "PERM_HINT" removing is fine
I am not sure we should remove PERM_HINT_GET_RATE.
if some old contract will set rate with this bit?
maybe we should keep turning it off??

contractsV5/KyberNetwork.sol
```diff
@@ -58,23 +55,14 @@ contract KyberNetwork is Withdrawable, Utils, IKyberNetwork, ReentrancyGuard {
...
-        tradeData.input.src = src;
-        tradeData.input.srcAmount = srcAmount;
-        tradeData.input.dest = dest;
-        tradeData.input.destAddress = destAddress;
-        tradeData.input.maxDestAmount = maxDestAmount;
-        tradeData.input.minConversionRate = minConversionRate;
-        tradeData.input.platformWallet = platformWallet;
-        tradeData.input.platformFeeBps = platformFeeBps;
-
+        initTradeData(trader, src, srcAmount, dest, destAddress, maxDestAmount, minConversionRate, platformWallet, platformFeeBps, tradeData);
```
ilanDoron: not sure why is this better?
and why here one liner and next one is one below other?
anyway, if keeping this long function call
please use safe call convention . with parameter names
i.e.

callFuction(paramName: input, nextParamName: input2 etc...

contractsV5/KyberNetwork.sol
```diff
@@ -106,8 +98,8 @@ contract KyberNetwork is Withdrawable, Utils, IKyberNetwork, ReentrancyGuard {
     /// @dev add or deletes a reserve to/from the network.
     /// @param reserve The reserve address.
     function addReserve(IKyberReserve reserve, uint reserveId, bool isFeePaying) public onlyOperator returns(bool) {
-        require(reserveIdToAddresses[reserveId][0] == address(0));
-        require(reserveAddressToId[address(reserve)] == uint(0));
+        require(reserveIdToAddresses[reserveId].length == 0, "reserve id already taken");
```
ilanDoron: current plan is that each reserve ID
has current reserve in cell 0
and others in next cells

I assume the remove reserve will remove the one in array cell 0 and move it to last array cell.
cell 0 address will be set to address(0)
this is why I make sure cell 0 with address(0)
so since its mapping array of addresses. length will be 0 only for first address per ID

contractsV5/KyberNetwork.sol
```diff
@@ -399,16 +404,17 @@ contract KyberNetwork is Withdrawable, Utils, IKyberNetwork, ReentrancyGuard {
...
             rates[i] = reserve.getConversionRate(
                 src,
                 dest,
                 srcAmountWithFee,
                 block.number);
 
-            destAmount = srcAmount * rates[i] / PRECISION;
-             //if token -> ETH swap & reserve is not feeless, deduct fee from destAmount
-            destAmount = (dest == ETH_TOKEN_ADDRESS && isPayingFees) ? destAmount * (BPS - takerFeeBps) / BPS : destAmount;
+            destAmount = srcAmountWithFee * rates[i] / PRECISION;
```
ilanDoron: this isn't destAmount
maybe call it rateReference?

contractsV5/BytesLib.sol
```diff
@@ -0,0 +1,506 @@
+/*
+ * @title Solidity Bytes Arrays Utils
+ * @author Gonçalo Sá <goncalo.sa@consensys.net>
+ *
+ * @dev Bytes tightly packed arrays utility library for ethereum contracts written in Solidity.
+ *      The library lets you concatenate, slice and type cast bytes arrays both in memory and storage.
```
ilanDoron: what type of license this lib has?
please check...

contractsV5/KyberHintParser.sol
```diff
@@ -0,0 +1,146 @@
...
+                }
+            }
+            require((bpsSoFar == BPS) || (bpsSoFar == 0), "bps do not sum to 10000");
+        }
+    }
+    
+    function encodeUint(uint num) public pure returns (bytes memory hint) {
+        return hint.concat(abi.encodePacked(uint16(num)));
+    }
+}
```
ilanDoron: CR LF

contractsV5/KyberNetwork.sol
```diff
@@ -85,18 +73,22 @@ contract KyberNetwork is Withdrawable, Utils, IKyberNetwork, ReentrancyGuard {
...
-        tradeData.input.trader = address(uint160(trader));
-        tradeData.input.src = src;
-        tradeData.input.srcAmount = srcAmount;
-        tradeData.input.dest = dest;
-        tradeData.input.destAddress = address(uint160(destAddress));
-        tradeData.input.maxDestAmount = maxDestAmount;
-        tradeData.input.minConversionRate = minConversionRate;
-        tradeData.input.platformWallet = walletId;
-        tradeData.input.platformFeeBps = 0;
+        initTradeData(
```
ilanDoron: actually the above is better. since its safer.

if using this one
please use the set parameter by name convention...
same for all initTradeData  calls.

contractsV5/KyberNetwork.sol
```diff
@@ -280,19 +272,25 @@ contract KyberNetwork is Withdrawable, Utils, IKyberNetwork, ReentrancyGuard {
...
-        require(expectedRateContract != IExpectedRate(0));
+        // require(expectedRateContract != IExpectedRate(0));
         if (src == dest) return (0, 0);
+        TradeData memory tradeData;
+        bytes memory hint;
 
-        uint qty = srcQty & ~PERM_HINT_GET_RATE;
-        
-        ( , , expectedRate) = expectedRateContract.getExpectedRate(src, dest, qty);
+        initTradeData(address(uint160(0)), src, srcQty, dest, address(uint160(0)), 2 ** 255, 0, address(uint160(0)), 0, tradeData);
```
ilanDoron: same comment as above. for initTradeData call

Anyhowclick:
Changes to BytesLib and KyberHintParser will be done by Anton.

contractsV5/KyberNetwork.sol
Anyhowclick: AFAIK, no one uses it....

I suppose we can, though it only means additional gas ¯\_(ツ)_/¯

contractsV5/BytesLib.sol
Anyhowclick: The Unlicense.

"A license with no conditions whatsoever which dedicates works to the public domain. Unlicensed works, modifications, and larger works may be distributed under different terms and without source code."

contractsV5/KyberNetwork.sol
Anyhowclick: Will swtich to the one below the other.

contractsV5/KyberNetwork.sol
Anyhowclick: It means duplicate code, but if that's safer, ok.

Anyhowclick:
@ayobuenavista kindly take note of the comments to BytesLib and the parser

contractsV5/KyberNetwork.sol
Anyhowclick: No, rateReference isn't the right term, since it's calculating an amount from the rate.

contractsV5/KyberNetwork.sol
Anyhowclick: 1) Abstraction, since it's being used by the trade and getExpectedRate functions. 
2) Just need to change the underlying the function, and it ensures all the relevant inputs are initialised. 
3) Run the risk of missing an initialisation of a param, or wrong initialisation.

Agree with the safe call convention if we're deciding to stick to it. I've reverted to as is for now.

contractsV5/IKyberHint.sol
```diff
@@ -19,7 +19,7 @@ interface IKyberHint {
 
     function parseTokenToTokenHint(bytes calldata hint) external view
     returns(HintType tokenToEthType, address[] memory tokenToEthReserves, uint[] memory tokenToEthSplits,
-        HintType ethToTokenType, address[] memory ethToTokenReserves, uint[] memory ethToTokenSplits);
+        HintType ethToTokenType, address[] memory ethToTokenReserves, uint[] memory ethToTokenSplits, uint failureHint);
```
Anyhowclick: I rebased my branch, but it's still showing this change for some reason.

contractsV5/IKyberHint.sol
```diff
@@ -12,22 +12,22 @@ interface IKyberHint {
     }
 
     function parseEthToTokenHint(bytes calldata hint) external view
-        returns(HintType hintType, address[] memory reserves, uint[] memory splits, uint failureHint);
+    returns(HintType hintType, address[] memory reserves, uint[] memory splits, uint failureHint);
```
Anyhowclick: Can we keep the same indentation standard among the interfaces? Compare the current IKyberHint and IKyberNetwork files and lemme know which to follow, thanks.

contractsV5/KyberNetwork.sol
ilanDoron: ok
reference amount then
its not dest amount

contractsV5/IKyberHint.sol
ilanDoron: I think per function the return value should be indented.

contractsV5/IKyberHint.sol
ilanDoron: yes
I want it indented.
what do u expect?

contractsV5/KyberNetwork.sol
```diff
@@ -97,25 +91,26 @@ contract KyberNetwork is Withdrawable, Utils, IKyberNetwork, ReentrancyGuard {
...
+    event AddReserveToNetwork(address indexed reserve, uint indexed reserveId, bool add, address indexed rebateWallet);
 
     /// @notice can be called only by operator
     /// @dev add or deletes a reserve to/from the network.
     /// @param reserve The reserve address.
-    function addReserve(IKyberReserve reserve, uint reserveId, bool isFeePaying, address wallet) public onlyOperator returns(bool) {
-        require(reserveIdToAddresses[reserveId][0] == address(0));
-        require(reserveAddressToId[address(reserve)] == uint(0));
+    function addReserve(address reserve, uint reserveId, bool isFeePaying, address wallet) public onlyOperator returns(bool) {
+        require(reserveIdToAddresses[reserveId].length == 0);
```
ilanDoron: this code is wrong
did u follow my explanation?
in previous PR?

PR #571, opened 2020-01-16
[WIP] separate trade logic to different contract

ilanDoron:


contractsV5/IKyberTradeLogic.sol
```diff
@@ -0,0 +1,42 @@
...
+         destAmountWithNetworkFee,
+         last
+    }
+    
+    enum FeesIndex {
+        takerFee,
+        customFee
+    }
+    
+    function addReserve(address reserve, uint reserveId, bool isFeePaying) external returns(bool);
```
Anyhowclick: I think `addReserve` and `listPairForReserve` should not be part of the KyberTradeLogic contract, since as its name suggests, should only have the trade logic.

contractsV5/KyberTradeLogic.sol
```diff
@@ -0,0 +1,414 @@
...
+            bestReserve.index = reserveCandidates[uint(blockhash(block.number-1)) % numRelevantReserves];
+        } else {
+            bestReserve.index = reserveCandidates[0];
+        }
+        
+        isPayingFees = isFeePayingReserve[address(reserveArr[bestReserve.index])];
+        return (reserveArr[bestReserve.index], rates[bestReserve.index], isPayingFees);
+    }
+
+    uint constant rateMul = 1;
```
Anyhowclick: What's this lol... Unused constant.

contractsV5/IKyberTradeLogic.sol
ilanDoron: trade logic must have this data for running trade logic.
I am thinking to network will call trade logic.
so we don't need a seperate call.
also notice API here is a bit different. un needed data isn't sent here.

regarding list pair
I am also thinking it should call this API (so we don't need separate call)
I also think listing of token per reserve will not be done in network. but only here.

contractsV5/KyberTradeLogic.sol
ilanDoron: ho yes.
So I wanted some multiplier to use for finding the rate in the per reserve array. which holds rate and split BPS
currently I mul by constant.
will probably move to this....

lets discuss monday

ilanDoron:
handled in another PR

PR #572, opened 2020-01-17
Make perm stuff local, refactor to initTradeData, findRatesAndAmounts -> calcRatesAndAmounts

Anyhowclick:


contractsV5/KyberNetwork.sol
```diff
@@ -100,12 +98,14 @@ contract KyberNetwork is Withdrawable, Utils, IKyberNetwork, ReentrancyGuard {
...
     /// @param reserve The reserve address.
     function addReserve(address reserve, uint reserveId, bool isFeePaying, address wallet) public onlyOperator returns(bool) {
-        require(reserveIdToAddresses[reserveId].length == 0);
-        require(reserveAddressToId[reserve] == uint(0));
+        require(reserveIdToAddresses[reserveId].length == 0, "reserveId points to existing reserve");
+        require(reserveAddressToId[reserve] == uint(0), "reserve has an existing id");
         
         reserveAddressToId[reserve] = reserveId;
 
+        reserveIdToAddresses[reserveId] = new address[](1);
```
ilanDoron: notice
this array is holding all reserves. also previous ones.
So it will get larger as reserves are added...

contractsV5/KyberNetwork.sol
```diff
@@ -266,20 +266,27 @@ contract KyberNetwork is Withdrawable, Utils, IKyberNetwork, ReentrancyGuard {
         returns (uint expectedRate, uint worstRate)
     {
         if (src == dest) return (0, 0);
+        uint permHintGetRate = 1 << 255; //for backwards compatibility
```
ilanDoron: I think constant is still preferred
otherwise this is calculated per function call.
constant is calculated on compilation time.
but I am not sure about this.
Should check the assembly...
or check gas consumption. 
or see documentation.
which ever is easier

contractsV5/KyberNetwork.sol
```diff
@@ -266,20 +266,27 @@ contract KyberNetwork is Withdrawable, Utils, IKyberNetwork, ReentrancyGuard {
...
+        uint qty = ((srcQty & permHintGetRate) > 0) ? srcQty & ~permHintGetRate : srcQty;
```
ilanDoron: I think just turn the bit off.
and avoid extra code.
unless it saves gas the way its don't here. which I don't see why it should.

contractsV5/KyberNetwork.sol
```diff
@@ -893,9 +930,16 @@ contract KyberNetwork is Withdrawable, Utils, IKyberNetwork, ReentrancyGuard {
...
     function parseTradeDataHint(TradeData memory tradeData,  bytes memory hint) internal view {
-        tradeData.tokenToEth.addresses = reservesPerTokenSrc[address(tradeData.input.src)];
-        tradeData.ethToToken.addresses = reservesPerTokenDest[address(tradeData.input.dest)];
-        //PERM_HINT is treated as no hint, so we just return
-        if (hint.length == 0 || keccak256(hint) == keccak256(PERM_HINT)) return;
+        tradeData.tokenToEth.addresses = (tradeData.input.src == ETH_TOKEN_ADDRESS) ? new IKyberReserve[](1) : reservesPerTokenSrc[address(tradeData.input.src)];
+        tradeData.ethToToken.addresses = (tradeData.input.dest == ETH_TOKEN_ADDRESS) ? new IKyberReserve[](1) :reservesPerTokenDest[address(tradeData.input.dest)];
+
+        //PERM is treated as no hint, so we just return
+        if (hint.length == 0 || keccak256(hint) == keccak256("PERM")) {
```
ilanDoron: for gas consumption
is it better to check perm only of length is 4.
I guess it should be better. please check.
so we don't run two keccak operations. per trade when not required.

contractsV5/KyberNetwork.sol
```diff
@@ -100,12 +98,14 @@ contract KyberNetwork is Withdrawable, Utils, IKyberNetwork, ReentrancyGuard {
     /// @dev add or deletes a reserve to/from the network.
     /// @param reserve The reserve address.
     function addReserve(address reserve, uint reserveId, bool isFeePaying, address wallet) public onlyOperator returns(bool) {
-        require(reserveIdToAddresses[reserveId].length == 0);
-        require(reserveAddressToId[reserve] == uint(0));
+        require(reserveIdToAddresses[reserveId].length == 0, "reserveId points to existing reserve");
```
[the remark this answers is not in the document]
ilanDoron: in general my idea is
when reserve removed, it will be placed in next array cell.
and cell 0 will be zeroed.
so the check should be cell 0 has zero address....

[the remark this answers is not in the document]
Anyhowclick: The problem is the first instantiation, when there is no array in the first place. Discovered it when I was writing the test code.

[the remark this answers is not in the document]
Anyhowclick: Edited this part

contractsV5/KyberNetwork.sol
Anyhowclick: Fixed

contractsV5/KyberNetwork.sol
Anyhowclick: 20 gas cheaper with global constant

contractsV5/KyberNetwork.sol
Anyhowclick: 20 gas cheaper with global constant

contractsV5/KyberNetwork.sol
Anyhowclick: It's 400 gas more to use keccak.

contractsV5/KyberNetwork.sol
```diff
@@ -48,17 +46,17 @@ contract KyberNetwork is Withdrawable, Utils, IKyberNetwork, ReentrancyGuard {
...
+            trader: trader,
+            src: src,
+            dest: dest,
+            srcAmount: srcAmount,
+            destAddress: destAddress,
+            maxDestAmount: maxDestAmount,
+            minConversionRate: minConversionRate,
+            platformWallet: platformWallet,
+            platformFeeBps: platformFeeBps
+            });
```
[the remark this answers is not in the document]
Anyhowclick: Let's not kill ourselves now lol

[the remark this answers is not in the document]
ilanDoron: yes yes
I know
we will see

its good task for intern...
:)

PR #575, opened 2020-01-20
Adding FeeHandler.sol and updating IFeeHandler and IKyberDAO interfaces.

STYJ:


contractsV5/FeeHandler.sol
```diff
@@ -0,0 +1,102 @@
...
+import "./IKyberDAO.sol";
+import "./KyberNetwork.sol";
+import "./IFeeHandler.sol";
+import "./PermissionGroupsV5.sol";
+import "./UtilsV5.sol";
+
+contract FeeHandler is IFeeHandler, Utils {
+
+    IKyberDAO public kyberDAOContract;
+    KyberNetwork public kyberNetworkContract;
```
ilanDoron: here we want the interface.
or address.

otherwise you include the whole network code and deploy it as part of your code.

contractsV5/FeeHandler.sol
```diff
@@ -0,0 +1,102 @@
...
+        // if we include a dest address, we need an owner / admin of the reserve and the below function.
+    }
+
+    function burnKNC() public {
+        // only DAO?
+        // convert fees to KNC and burn
+        // Eth for burning is the remaining == (total balance - total_reward_amount - total_reserve_rebate).
+    }
+
+    function setReserveRebateWallet(address reserve, address wallet) public returns (bool){}
```
ilanDoron: we will not set rebate wallet here.
we receive an array of rebate wallets and do accounting.
since we don't have admin to set rebate wallets.

contractsV5/FeeHandler.sol
```diff
@@ -0,0 +1,102 @@
...
+        // only DAO
+        // send rebate to reserve
+        // update rebatePerReserve;
+        // update total rebate amounts?
+        // update reserve rebate to 1 (avoid 0...) otherwise div by 0 issue? but will we even need to div by 0?
+        // if we include a dest address, we need an owner / admin of the reserve and the below function.
+    }
+
+    function burnKNC() public {
+        // only DAO?
```
ilanDoron: anyone can burn.

contractsV5/FeeHandler.sol
```diff
@@ -0,0 +1,102 @@
...
+
+    function claimStakerReward(address staker, uint percentageinPrecision, uint epoch) public {
+        // onlyDAO
+        // send reward
+        // update rewardPerEpoch
+        // update totalReward
+    }
+
+    function claimReserveRebate(address reserve) public {
+        // only DAO
```
ilanDoron: anyone can claim
will send amounts to the rebate wallet address if it has any values saved in array....

please make sure
Always leave 1 wei here.
so we don't zero the value
this to avoid high gas cost of changing value for 0 to non zero

here think on the attack vector....

contractsV5/FeeHandler.sol
```diff
@@ -0,0 +1,102 @@
...
+            totalRewardsPerEpoch[epoch] += rewardWei;
+            // Internal accounting for total rewards (totalRewards)
+            totalRewards += rewardWei;
+        }
+        // Todo: emit event
+        return true;
+    }
+
+
+    function claimStakerReward(address staker, uint percentageinPrecision, uint epoch) public {
```
ilanDoron: percentageinPrecision

capitalize i
percentageInPrecision

contractsV5/FeeHandler.sol
```diff
@@ -0,0 +1,102 @@
...
+import "./PermissionGroupsV5.sol";
+import "./UtilsV5.sol";
+
+contract FeeHandler is IFeeHandler, Utils {
+
+    IKyberDAO public kyberDAOContract;
+    KyberNetwork public kyberNetworkContract;
+
+    // Todo: Add the correct startBlock and epoch duration values
+    uint constant STARTBLOCK = 0;
```
ilanDoron: we will not save those values.

the expiry block parameter saves the need to save locally.
so only DAO knows about epoch number and duration.
so can remove these values

contractsV5/FeeHandler.sol
STYJ: I need to call the variable `reserveRebateWallets` that does not exist in the interface.

contractsV5/FeeHandler.sol
STYJ: Right, maybe it's more appropriate to scope it within KyberNetwork.sol? I'm just thinking about the case where the reserveRebateWallet was not set when the reserve is added or if they decided to change it in the future. 

Also, don't worry i'm not spending much time on it that's why I haven't implemented it!

contractsV5/FeeHandler.sol
STYJ: yes anyone can call the burnKNC function from the DAO but the burnKNC function from the FeeHandler can only be called by the DAO

STYJ: I suppose it doesn't hurt to remove the modifier.

contractsV5/FeeHandler.sol
STYJ: Okay I understand what you mean now. We can declare a function in the interface for the the public variable like the `decimals()` function in the ERC20 interface. 

Do we want to add this into the `IKyberNetwork.sol` or a separate interface?

contractsV5/FeeHandler.sol
STYJ: fixed

contractsV5/FeeHandler.sol
STYJ: Mike mentioned in the group (https://t.me/c/1407851547/3537) that I can hardcode `STARTBLOCK` and `EPOCHDURATION` in the FeeHandler. I can remove it if it is not needed. I don't think I need it.

contractsV5/FeeHandler.sol
STYJ: I feel like just to be safe, it is better to have an onlyDAO modifier but at the DAO contract, anyone can call it... 

Noted on your other points.

contractsV5/FeeHandler.sol
ilanDoron: why does fee handle need to call reserveRebateWallets? its getting the wallet list.
its doesn't know about the reserves at all. only the wallets.

contractsV5/FeeHandler.sol
ilanDoron: the fee handler will get a list of rebate wallets. and update their accounts with internal accounting.

if the reserve updates a wallet in the future. fee handler doesn't need to know about it.
any one who updated his wallet will have to know he should claim for different wallets.

contractsV5/FeeHandler.sol
ilanDoron: I think the burning isn't related to the DAO
DAO only sets BRR values.
but burning is done by anyone.

we will have to limit amounts. per burn.
and create some block delay between burns.

contractsV5/FeeHandler.sol
ilanDoron: we don't want to add code to the DAO if not necessary.
and if anyone can do it from the DAO
why is it safer?

contractsV5/FeeHandler.sol
ilanDoron: if its fixed push it?
so we can see it here

contractsV5/FeeHandler.sol
ilanDoron: yes.
but later I asked him to decouple the logic.

its much better if the epoch logic is only in one place
the more we decouple logic - its better design

if later epoch logic changes.
fee handler doesn't have to be modified.

contractsV5/FeeHandler.sol
STYJ: It needs to know the `reserveRebateWallet` because in the `claimReserveRebate` function, the parameter passed into the function is a reserve address, not a rebate wallet address. I'm planning to change it to a rebate wallet address instead (assuming that the DAO has access to the KyberNetwork contract)

contractsV5/FeeHandler.sol
STYJ: noted

contractsV5/FeeHandler.sol
STYJ: noted

contractsV5/FeeHandler.sol
STYJ: I'm not adding any code to the DAO. I was proposing to add the `onlyDAO` modifier to the FeeHandler to ensure that only the DAO contract can interact with some of these functions but anyone can call the function if they interact with the DAO contract. 

I can't think of a reason why it's safer but just thought that we should limit the possible interactions in order to prevent potential misuse of the function that we didnt think of...

contractsV5/FeeHandler.sol
STYJ: I was waiting for your other comments to fix more before pushing but okay, will just push what I have.

contractsV5/FeeHandler.sol
STYJ: noted

contractsV5/FeeHandler.sol
STYJ: I've changed it to rebateWallet instead of using reserve address.

ilanDoron: so please remove the include for kyberNetwork
and also see if u can define kyber network as address parameter.

contractsV5/FeeHandler.sol
```diff
@@ -0,0 +1,189 @@
+pragma solidity 0.5.11;
+
+import "./IKyberDAO.sol";
+import "./IFeeHandler.sol";
+import "./PermissionGroupsV5.sol";
+import "./KyberNetworkProxy.sol";
```
ilanDoron: fee handler shouldn't include kyberNetworkProxy code. 
use interface instead.

contractsV5/FeeHandler.sol
```diff
@@ -0,0 +1,189 @@
...
+
+    constructor(
+        IKyberDAO _kyberDAOContract,
+        KyberNetworkProxy _kyberNetworkProxyContract,
+        address _knc,
+        address payable _burnAddress,
+        uint _burnBlockInterval
+    ) public
+    {
+        kyberDAOContract = _kyberDAOContract;
```
ilanDoron: please require non zero addresses

contractsV5/FeeHandler.sol
```diff
@@ -0,0 +1,189 @@
...
+        require(
+            msg.sender == address(kyberDAOContract),
+            "Only the DAO can call this function."
+        );
+        _;
+    }
+
+    modifier onlyKyberNetwork {
+        require(
+            msg.sender == address(kyberNetworkProxyContract.kyberNetworkContract()),
```
ilanDoron: we don't want external function call to this modifier. its expensive. 
please accept network contract address in ctor.
and set it locally (storage) as address (don't need interface.)

contractsV5/FeeHandler.sol
```diff
@@ -0,0 +1,189 @@
...
+            "Only the internal KyberNetwork contract can call this function."
+        );
+        _;
+    }
+
```
ilanDoron: remove space

contractsV5/FeeHandler.sol
```diff
@@ -0,0 +1,189 @@
...
+
+    function encodeData(uint _burn, uint _reward, uint _epoch, uint _expiryBlock) public pure returns (uint) {
+        return (((((_burn << BITS_PER_PARAM) + _reward) << BITS_PER_PARAM) + _epoch) << BITS_PER_PARAM) + _expiryBlock;
```
ilanDoron: nice functions. (encode, decode)
but remove them down in the file.
main logic better be on top.

these are like helper functions.

contractsV5/FeeHandler.sol
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    event AccumulateReserveRebate(address rebateWallet, uint rebateAmountWei);
+
+    // Todo: future optimisation to accumulate rebates for 2 rebate wallet
+    // encode totals, 128 bits per reward / rebate
+    function handleFees(address[] calldata eligibleWallets, uint[] calldata rebatePercentages) external payable onlyKyberNetwork returns(bool) {
+        // Decoding BRR data
+        (uint burnInBPS, uint rewardInBPS, uint epoch, uint expiryBlock) = decodeData();
+        uint rebateInBPS = BPS - rewardInBPS - burnInBPS;
```
ilanDoron: I thought we decided burn will be calculated.

contractsV5/FeeHandler.sol
```diff
@@ -0,0 +1,189 @@
...
+            brrAndEpochData = encodeData(burnInBPS, rewardInBPS, epoch, expiryBlock);
+        }
+
+        uint rebateWei = rebateInBPS * msg.value / BPS;
+        uint rewardWei = rewardInBPS * msg.value / BPS;
+        for(uint i = 0; i < eligibleWallets.length; i ++) {
+            // Internal accounting for rebates per reserve wallet (totalRebatesPerRebateWallet)
+            totalRebatesPerRebateWallet[eligibleWallets[i]] = rebateWei * rebatePercentages[i] / 100;
+            // Internal accounting for total rebates (totalRebates)
+            totalRebates += rebateWei * rebatePercentages[i] / 100;
```
ilanDoron: please do this outside the loop for the full value.
save storage updates!

contractsV5/FeeHandler.sol
```diff
@@ -0,0 +1,189 @@
...
+            // Internal accounting for reward per epoch (totalRewardsPerEpoch)
+            totalRewardsPerEpoch[epoch] += rewardWei;
```
ilanDoron: do this outside the loop

inside loop will result in wrong number. no?

contractsV5/FeeHandler.sol
```diff
@@ -0,0 +1,189 @@
...
+            // Internal accounting for total rewards (totalRewards)
+            totalRewards += rewardWei;
+
+            emit AccumulateReserveRebate(eligibleWallets[i], rebateWei * rebatePercentages[i] / 100);
```
ilanDoron: better one event for all the data. including wallet array. wallet rebates. full rebate amount. full reward amount. full burn amount.

contractsV5/FeeHandler.sol
```diff
@@ -0,0 +1,189 @@
...
+        uint rewardWei = rewardInBPS * msg.value / BPS;
+        for(uint i = 0; i < eligibleWallets.length; i ++) {
+            // Internal accounting for rebates per reserve wallet (totalRebatesPerRebateWallet)
+            totalRebatesPerRebateWallet[eligibleWallets[i]] = rebateWei * rebatePercentages[i] / 100;
+            // Internal accounting for total rebates (totalRebates)
+            totalRebates += rebateWei * rebatePercentages[i] / 100;
+            // Internal accounting for reward per epoch (totalRewardsPerEpoch)
+            totalRewardsPerEpoch[epoch] += rewardWei;
+            // Internal accounting for total rewards (totalRewards)
+            totalRewards += rewardWei;
```
ilanDoron: move outside loop

contractsV5/FeeHandler.sol
```diff
@@ -0,0 +1,189 @@
...
+            emit AccumulateReserveRebate(eligibleWallets[i], rebateWei * rebatePercentages[i] / 100);
+        }
+        return true;
+    }
+
+    event DistributeRewards(address staker, uint amountWei);
+
+    function claimStakerReward(address staker, uint percentageInPrecision, uint epoch) public onlyDAO returns(uint) {
+        // Amount of reward to be sent to staker
+        uint amount = totalRewardsPerEpoch[epoch] * percentageInPrecision / 10**18;
```
ilanDoron: in utils we should have PRECISION constant.

contractsV5/FeeHandler.sol
```diff
@@ -0,0 +1,189 @@
...
+        totalRewards -= amount;
+
+        // send reward to staker
+        (bool success, ) = staker.call.value(amount)("");
+        require(success, "Transfer of rewards to staker failed.");
+
+        emit DistributeRewards(staker, amount);
+
+        return amount;
+
```
ilanDoron: remove space please

contractsV5/FeeHandler.sol
```diff
@@ -0,0 +1,189 @@
...
+    event DistributeRebate(address rebateWallet, uint amountWei);
+
+    // Using rebateWallet instead of reserve so I don't have to store KyberNetwork variable.
+    function claimReserveRebate(address rebateWallet) public returns (uint){
+        // Get total amount of rebate accumulated
+        uint amount = totalRebatesPerRebateWallet[rebateWallet] - 1;
+
+        // Update total rebate and rebate per rebate wallet amounts
+        totalRebates -= amount;
+        totalRebatesPerRebateWallet[rebateWallet] = 1; // Do not use 0 to avoid potential issues with div by 0
```
ilanDoron: I was thinking not to use 0 to avoid paying high gas when setting back from 0 to non zero.

contractsV5/FeeHandler.sol
```diff
@@ -0,0 +1,189 @@
...
+    // we will have to limit amounts. per burn.
+    // and create some block delay between burns.
+    function burnKNC() public returns(uint) {
+        // check if current block > last buy block number + num block interval
+        require(
+            block.number < lastBurnBlock + burnBlockInterval,
+            "Unable to burn as burnBlockInterval has not passed since lastBurnBlock"
+        );
+
+        // update last buy block number
```
ilanDoron: typo
burn

contractsV5/FeeHandler.sol
```diff
@@ -0,0 +1,189 @@
...
+        lastBurnBlock = block.number;
+
+        // Get srcQty to burn, if greater than 10 ETH, burn only 10 ETH per function call.
```
ilanDoron: don't use number (10) in comment.
use the constant name.

contractsV5/FeeHandler.sol
```diff
@@ -0,0 +1,189 @@
...
+        uint srcQty = address(this).balance - totalRebates - totalRewards;
+
+        srcQty = srcQty > ETH_TO_BURN ? ETH_TO_BURN : srcQty;
+
+        // Get the slippage rate
+        // If srcQty is too big, get expected rate will return 0 so maybe we should limit how much can be bought at one time.
```
ilanDoron: I think we should set around 5 eth. or maybe less
but lets think about it more.
we also want to check there is no arbitrage before moving forward.

contractsV5/FeeHandler.sol
```diff
@@ -0,0 +1,189 @@
...
+            srcQty,
+            IERC20(knc),
+            burnAddress,
+            MAX_QTY,
+            expectedRate * 97 / 100,
+            address(0), // platform wallet
+            0, // platformFeeBps
+            "" // hint
+        );
+
```
ilanDoron: regarding burn.
knc has burn function.
use it.
don't need the burn address.
See how its done in fee burner contract.

contractsV5/FeeHandler.sol
STYJ: noted

contractsV5/FeeHandler.sol
STYJ: noted

contractsV5/FeeHandler.sol
STYJ: noted

contractsV5/FeeHandler.sol
STYJ: noted and updated, will push when the other comments are fixed.

contractsV5/FeeHandler.sol
STYJ: noted and updated, will push when the other comments are fixed.

contractsV5/FeeHandler.sol
STYJ: It's the same either way... I can change it to burn if you want. Just need to change the encode decode function

contractsV5/FeeHandler.sol
STYJ: Ah yes, I missed this out. will change and push when the other comments are fixed.

contractsV5/FeeHandler.sol
STYJ: Yes, wrong logic, will fix and push together with the other fixes

contractsV5/FeeHandler.sol
STYJ: updated and will push fix when other comments are fixed

contractsV5/FeeHandler.sol
STYJ: yes, wrong logic, will move outside of loop and push fix when other comments are fixed

contractsV5/FeeHandler.sol
STYJ: updated and will push fix when other comments are fixed

contractsV5/FeeHandler.sol
STYJ: noted, updated and will push fix when other comments are fixed

contractsV5/FeeHandler.sol
STYJ: noted, updated and will push fix when other comments are fixed

contractsV5/FeeHandler.sol
STYJ: noted, updated and will push fix when other comments are fixed

contractsV5/FeeHandler.sol
STYJ: noted, updated and will push fix when other comments are fixed

contractsV5/FeeHandler.sol
STYJ: Ok, I'm just using 10 eth as a temporary value for now.

contractsV5/FeeHandler.sol
STYJ: As discussed in our call, I will change this to calculate burn

contractsV5/FeeHandler.sol
STYJ: The `FeeBurner` contract calls `knc.burnFrom` which in turn calls `transferFrom` and then `burn` function. Since the knc is already in the FeeHandler contract, I can just call the `burn` function. Will add a IBurnableToken interface.

ilanDoron: yes.
if it works I guess its better.

contractsV5/IKyberDAO.sol
```diff
@@ -0,0 +1,9 @@
+pragma solidity 0.5.11;
+
+interface IKyberDAO {
+    function getLatestNetworkFeeData() external view returns(uint feeInBps, uint expiryBlockNumber);
+    function getLatestBRRData() external view returns(uint burnInBps, uint rewardInBps, uint rebateInBps, uint epoch, uint expiryBlockNumber);
+    function claimStakerReward(uint epoch) external returns(uint);
+    function claimReserveRebate(uint epoch) external returns(uint);
+    function burnKNC() external returns(uint);
```
ilanDoron: I don't think DAO needs to know about burn KNC
it will be public in fee handler for anyone to call.

contractsV5/FeeHandler.sol
```diff
@@ -0,0 +1,200 @@
...
+import "./PermissionGroupsV5.sol";
+import "./IKyberNetworkProxy.sol";
+import "./UtilsV5.sol";
+import "./IERC20.sol";
+import "./IBurnableToken.sol";
+
+contract FeeHandler is IFeeHandler, Utils {
+
+    IKyberDAO public kyberDAOContract;
+    // if I use the interface, I am not able to get the KyberNetwork internal contract from the proxy..
```
ilanDoron: I get your comment
the correct solution to this is create a local interface that can do this. and not include all the code.
Anyway now its set in ctor. so no problem.

contractsV5/FeeHandler.sol
```diff
@@ -0,0 +1,200 @@
...
+    IKyberNetworkProxy public kyberNetworkProxyContract;
+    address public kyberNetworkContract;
+    address public knc;
+    uint public burnBlockInterval;
+    uint public lastBurnBlock;
+    uint constant ETH_TO_BURN = 10**19;
```
ilanDoron: please place constants together and dynamic together.
I think the convention is constant first. but not sure.

contractsV5/FeeHandler.sol
```diff
@@ -0,0 +1,200 @@
...
+            address(this),
+            MAX_QTY,
+            expectedRate * 97 / 100,
+            address(0), // platform wallet
+            0, // platformFeeBps
+            "" // hint
+        );
+
+        // Burn KNC
+        require(IBurnableToken(knc).burn(destQty), "KNC burn failed");
```
ilanDoron: can define knc as burnable token in the first place. so don't have to cast it.

contractsV5/IKyberDAO.sol
STYJ: okay, removing

contractsV5/FeeHandler.sol
STYJ: yep, will remove comment.

contractsV5/FeeHandler.sol
STYJ: noted and updated

contractsV5/FeeHandler.sol
STYJ: noted and updated

contractsV5/FeeHandler.sol
ilanDoron: typo still here.

contractsV5/mock/MockDAO.sol
```diff
@@ -0,0 +1,39 @@
...
+        rewardInBPS = _rewardInBPS;
+        rebateInBPS = _rebateInBPS;
+        epoch = _epoch;
+        expiryBlockNumber = _expiryBlockNumber;
+    }
+
+    function setMockValues(uint _rewardInBPS, uint _rebateInBPS, uint _epoch, uint _expiryBlockNumber) public {
+        rewardInBPS = _rewardInBPS;
+        rebateInBPS = _rebateInBPS;
+        epoch = _epoch;
```
ilanDoron: better have separate API for setting epoch and expiry block.
notice epoch also used for network fee....

contractsV5/mock/MockDAO.sol
```diff
@@ -0,0 +1,39 @@
...
+        return (1, expiryBlockNumber);
+    }
+    function getLatestBRRData() external view returns(uint, uint, uint, uint, uint) {
+        return (BPS - rewardInBPS - rebateInBPS, rewardInBPS, rebateInBPS, epoch, expiryBlockNumber);
+    }
+    function claimStakerReward(uint _epoch) external returns(uint) {
+        return 1;
+    }
+    function claimReserveRebate(uint _epoch) external returns(uint) {
+        return 1;
```
ilanDoron: claim reserve rebate will not pass through the DAO

contractsV5/mock/MockDAO.sol
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    function getLatestNetworkFeeData() external view returns(uint, uint) {
+        return (1, expiryBlockNumber);
+    }
+    function getLatestBRRData() external view returns(uint, uint, uint, uint, uint) {
+        return (BPS - rewardInBPS - rebateInBPS, rewardInBPS, rebateInBPS, epoch, expiryBlockNumber);
+    }
+    function claimStakerReward(uint _epoch) external returns(uint) {
+        return 1;
```
ilanDoron: this should be calling fee handler with some precision units value.
maybe have the precision units an input parameter

test/v5/testFeeHandler.js
```diff
@@ -0,0 +1,54 @@
...
+
+
+contract('FeeHandler', function(accounts) {
+    before("Setting global variables", async() => {
+        kyberNetwork = accounts[0];
+        user = accounts[0];
+        kyberNetworkProxy = accounts[1];
+        rewardInBPS = new BN(3000);
+        rebateInBPS = new BN(5000);
+        epoch = new BN(0);
```
ilanDoron: epoch and expiry block will probably update as test proceeds.
so before each could update it to current block  + x
or something like this.

test/v5/testFeeHandler.js
```diff
@@ -0,0 +1,54 @@
...
+        Helper.assertEqual(expectedEncodedData, actualEncodedData, "Actual encoded data is not correct");
+    });
+    // Test decode
+
+    // Test handleFees
+    // Test claimStakerReward
+    // Test claimReserveRebate
+
+
+})
```
ilanDoron: add line

contractsV5/IBurnableToken.sol
```diff
@@ -0,0 +1,5 @@
+pragma solidity 0.5.11;
+
+interface IBurnableToken {
+    function burn(uint _value) external returns(bool);
+}
```
ilanDoron: add extra line here.

contractsV5/FeeHandler.sol
STYJ: sorry I don't see it. https://github.com/KyberNetwork/smart-contracts/pull/575/files#diff-56a4bcbd2878640347a559aef3fe0a8e

contractsV5/mock/MockDAO.sol
STYJ: done

test/v5/testFeeHandler.js
STYJ: done

contractsV5/IBurnableToken.sol
STYJ: done

test/v5/testFeeHandler.js
STYJ: noted, added an empty beforeEach but have not used it yet... I will implement it when i need to use it otherwise for now, I will keep my test simple.

contractsV5/mock/MockDAO.sol
STYJ: Understood, I got this from either yourself or Mike. Will remove from IKyberDAO interface.

contractsV5/mock/MockDAO.sol
STYJ: noted and added.

contractsV5/FeeHandler.sol
```diff
@@ -1,9 +1,205 @@
...
+        uint totalRewardAmtWei,
+        uint totalBurnAmtWei
+    );
+
+    // Todo: future optimisation to accumulate rebates for 2 rebate wallet
+    // encode totals, 128 bits per reward / rebate
+    function handleFees(address[] calldata eligibleWallets, uint[] calldata rebatePercentages) external payable onlyKyberNetwork returns(bool) {
+        // Decoding BRR data
+        (uint rewardInBPS, uint rebateInBPS, uint epoch, uint expiryBlock) = decodeData();
+        uint burnInBPS = BPS - rewardInBPS - rebateInBPS;
```
ilanDoron: burn In bps is actually not used.
can remove it from this function all together.
but in next PR please.

contractsV5/FeeHandler.sol
```diff
@@ -1,9 +1,205 @@
...
+    }
+
+    event DistributeRewards(address staker, uint amountWei);
+
+    function claimStakerReward(address staker, uint percentageInPrecision, uint epoch) public onlyDAO returns(uint) {
+        // Amount of reward to be sent to staker
+        uint amount = totalRewardsPerEpoch[epoch] * percentageInPrecision / PRECISION;
+
+        // Update total rewards and total rewards per epoch
+        totalRewardsPerEpoch[epoch] -= amount;
```
ilanDoron: need to make sure this isn't under flowing

contractsV5/FeeHandler.sol
ilanDoron: I see sentence
// update last buy block number

should be
// update last burn block number

contractsV5/FeeHandler.sol
```diff
@@ -1,9 +1,205 @@
...
+        );
+
+        // update last buy block number
+        lastBurnBlock = block.number;
+
+        // Get srcQty to burn, if greater than ETH_TO_BURN, burn only ETH_TO_BURN per function call.
+        uint srcQty = address(this).balance - totalRebates - totalRewards;
+        srcQty = srcQty > ETH_TO_BURN ? ETH_TO_BURN : srcQty;
+
+        // Get the slippage rate
```
ilanDoron: this isn't slippage rate. its rate.

test/v5/testFeeHandler.js
ilanDoron: sure

contractsV5/FeeHandler.sol
STYJ: done

contractsV5/FeeHandler.sol
STYJ: done

contractsV5/FeeHandler.sol
STYJ: done

PR #576, opened 2020-01-20
fix function signature for setImbalanceStepFunction

ilanDoron:


contracts/reserves/fprConversionRate/ConversionRateEnhancedSteps.sol
```diff
@@ -290,4 +290,4 @@ contract EnhancedStepFunctions is ConversionRates {
         if (y == 0) return false;
         return (((x*y) / y) != x);
     }
-}
+}
```
manhlx3006: Please fix this one

PR #586, opened 2020-01-21
Ilan/another npm fix

ilanDoron:


package.json
```diff
@@ -37,7 +37,7 @@
     "yargs": "11.1.1"
   },
   "devDependencies": {
-    "@nomiclabs/buidler": "1.0.2",
+    "@nomiclabs/buidler": "^1.0.2",
```
Anyhowclick: Eh, I thought u wanted to use specific version?

ilanDoron: yes
this PR not for real
only checking something else...
closing it

PR #598, opened 2020-01-28
update revert strings, add handling for taker fee updates in network …

ilanDoron:
…contract, add functions to save gas on network

### DIFF
--- a/contractsV5/KyberNetwork.sol
+++ b/contractsV5/KyberNetwork.sol
@@ -575,7 +575,7 @@ contract KyberNetwork is Withdrawable, Utils, IKyberNetwork, ReentrancyGuard {
         // uint rateWithAllFees;
     }
 
-  // accumulate fee wei
+    // accumulate fee wei
     function calcRatesAndAmounts(IERC20 src, IERC20 dest, uint srcAmount, TradeData memory tradeData) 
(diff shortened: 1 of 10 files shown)


contractsV5/KyberNetwork.sol
```diff
@@ -197,13 +198,15 @@ contract KyberNetwork is Withdrawable, Utils, IKyberNetwork, ReentrancyGuard {
         return true;
     }
 
-    event FeeHandlerContractSet(IFeeHandler newContract, IFeeHandler currentContract);
+    event ContractsUpdated(IFeeHandler newContract, IFeeHandler currentContract, IKyberDAO oldDAO, IKyberDAO newDAO);
```
Anyhowclick: Keep the order consistent?
So either old contract first, then new, or new first then old, ie.
1) `IFeeHandler oldFee, IFeeHandler newFee, IKyberDAO oldDAO, IKyberDAO newDAO`
2) `IFeeHandler newFee, IFeeHandler oldFee, IKyberDAO newDAO, IKyberDAO oldDAO`

ilanDoron: OK
changed the logic a bit.
anyway its fixed.

PR #601, opened 2020-01-30
Fix eth2dai set contracts func, add more tests

manhlx3006:


contractsV5/bridges/eth2dai/Eth2DaiReserve.sol
```diff
@@ -237,7 +237,12 @@ contract Eth2DaiReserve is IKyberReserve, Withdrawable, Utils {
         require(_otc != address(0), "setContracts: otc's address is missing");
 
         kyberNetwork = _kyberNetwork;
-        otc = IOtc(_otc);
+
+        if (_otc != address(otc)) {
+            // new otc address
+            otc = IOtc(_otc);
+            require(wethToken.approve(_otc, 2**255), "setContracts: failed to approve otc (wethToken)");
```
ilanDoron: should approve 0 to old OTC?

manhlx3006: Yeah, added reset allowance here

PR #611, opened 2020-02-02
remove local fee data from proxy

ilanDoron:


ilanDoron:
will be handled in another PR

PR #613, opened 2020-02-03
Anton/hint trade logic

ayobuenavista:


contractsV5/KyberHintHandler.sol
```diff
@@ -216,84 +247,103 @@ contract KyberHintParser is IKyberHint, Utils {
...
-            (reserves.addresses, reserves.splitValuesBps, indexToContinueFrom) = decodeReservesFromHint(false, hint, indexToContinueFrom);
-            decodeOperation(hint, tradeHint, indexToContinueFrom, isTokenToEth);
-        } else if (opcodeHash == keccak256(parseOpcode(TradeType.Split))) {
-            reserves.tradeType = TradeType.Split;
-            (reserves.addresses, reserves.splitValuesBps, indexToContinueFrom) = decodeReservesFromHint(true, hint, indexToContinueFrom);
-            decodeOperation(hint, tradeHint, indexToContinueFrom, isTokenToEth);
-        } else {
-            revert("Invalid hint opcode");
+            decodeOperation(hint, tradeHint, indexToContinueFrom, false, isExternal);
+        } else if (opcodeHash == keccak256(encodeOpcode(TradeType.MaskIn))) {
```
ilanDoron: does this method have better gas consumption then string compare?
we can consider having value of keccak256(encodeOpcode(TradeType.MaskIn)) ready as constant. and save run time.
unless compiler is smart enough to do it...
please check.

contractsV5/KyberHintHandler.sol
```diff
@@ -216,84 +247,103 @@ contract KyberHintParser is IKyberHint, Utils {
...
         uint indexToContinueFrom,
-        bool isTokenToEth
+        bool isTokenToEth,
+        bool isExternal
     )
         internal
         view
     {
         bytes memory opcode = hint.slice(indexToContinueFrom, 1);
         bytes32 opcodeHash = keccak256(opcode);
```
ilanDoron: I am not sure I understand why using keccak256?
is it for saving gas?

anyway.
please make sure the opcode compare is most gas efficient.
I am not sure it is right now.

What would be gas consumption if just take opcode

and later do: if(opcode == MASK_IN_OPCODE)
instead of keccak256 for both?
seems to me extra code is running here.
unless u found that this method is more gas efficient.
see comment below on same issue.

contractsV5/KyberHintHandler.sol
```diff
@@ -216,84 +247,103 @@ contract KyberHintParser is IKyberHint, Utils {
...
        }
+
        indexToContinueFrom++;
 
-       for (uint i = 0; i < reservesLength; i++) {           
-           addresses[i] = IKyberReserve(convertReserveIdToAddress(hint.slice(indexToContinueFrom, RESERVE_ID_LENGTH).toBytes8(0)));
+       for (uint i = 0; i < reservesLength; i++) {
+           reserveIds[i] = hint.slice(indexToContinueFrom, RESERVE_ID_LENGTH).toBytes8(0);
+
+           if (!isExternal) {
```
ilanDoron: since parsing external is running off chain.
this code shouldn't be here
it should be on the external call
we don't care to have another loop that translates all addresses to IDs.

contractsV5/KyberHintHandler.sol
```diff
@@ -216,84 +247,103 @@ contract KyberHintParser is IKyberHint, Utils {
...
            splitValuesBps = new uint[](1);
            splitValuesBps[0] = BPS;
        }
+
        indexToContinueFrom++;
 
-       for (uint i = 0; i < reservesLength; i++) {           
-           addresses[i] = IKyberReserve(convertReserveIdToAddress(hint.slice(indexToContinueFrom, RESERVE_ID_LENGTH).toBytes8(0)));
+       for (uint i = 0; i < reservesLength; i++) {
+           reserveIds[i] = hint.slice(indexToContinueFrom, RESERVE_ID_LENGTH).toBytes8(0);
```
ilanDoron: this code isn't required for internal call.
and this internal call is part of trade flow.

the default run should be only saving addresses.
and be gas efficient. 
And translating to IDs should be part of the off chain function that returns values to users.
we don't care to have another loop in the off chain view function.

contractsV5/KyberHintHandler.sol
```diff
@@ -216,84 +247,103 @@ contract KyberHintParser is IKyberHint, Utils {
...
                bpsSoFar += splitValuesBps[i];
                indexToContinueFrom += 2;
            }
        }
+
        require((bpsSoFar == BPS) || (bpsSoFar == 0), "bps do not sum to 10000");
-       return (addresses, splitValuesBps, indexToContinueFrom);
+
+       reservesHint.splitValuesBps = splitValuesBps;
+       if (isExternal) {
```
ilanDoron: also here please only have relevant code as part of trade flow.
we are talking about users gas.
we don't run unnecessary code per trade

contractsV5/KyberHintHandler.sol
```diff
@@ -216,84 +247,103 @@ contract KyberHintParser is IKyberHint, Utils {
...
+           reservesHint.reserveIds = reserveIds;
+       } else {
+           reservesHint.addresses = addresses;
+       }
+
+       return (indexToContinueFrom);
     }
 
-    function parseOpcode(TradeType h) internal pure returns (bytes memory) {
+    function encodeOpcode(TradeType h) internal pure returns (bytes memory) {
```
ilanDoron: what is h?
can it be type or something of that sort.

package.json
```diff
@@ -20,15 +20,15 @@
   },
   "homepage": "https://github.com/KyberNetwork/smart-contracts#readme",
   "dependencies": {
-    "@nomiclabs/buidler-truffle5": "1.0.2",
+    "@nomiclabs/buidler-truffle5": "^1.1.2",
```
ilanDoron: this is for error logs?
anyway
usually its better to commit web3 updates in separate commits.
and its better to do it for main branch.
since we will later do the same PR for development and Master....

so please don't add it here
rather do it in development branch.

also remove the ^
we want to know exactly which package we use.
And we want to be consistent across all runs.

package.json
```diff
@@ -20,15 +20,15 @@
...
     "chai-as-promised": "7.1.1",
     "compare-versions": "3.5.1",
     "ganache-cli": "6.8.2",
     "mathjs": "4.4.2",
     "rlp": "2.2.4",
     "secp256k1": "3.8.0",
     "sha3": "2.1.1",
-    "solc": "0.4.18",
+    "solc": "^0.5.11",
```
ilanDoron: do this in development branch
without the ^

package.json
```diff
@@ -37,9 +37,9 @@
     "yargs": "11.1.1"
   },
   "devDependencies": {
-    "@nomiclabs/buidler": "1.0.2",
-    "@nomiclabs/buidler-web3": "1.0.2",
-    "@openzeppelin/test-helpers": "0.5.4",
+    "@nomiclabs/buidler": "^1.1.2",
+    "@nomiclabs/buidler-web3": "^1.1.2",
+    "@openzeppelin/test-helpers": "^0.5.4",
```
ilanDoron: same comments as above.

package.json
ayobuenavista: Yes, for error logging and also running buidler tests. This commit already removed from the commit tree.

package.json
ayobuenavista: Removed from commit tree.

package.json
ayobuenavista: Removed from commit tree.

contractsV5/KyberHintHandler.sol
ayobuenavista: Will change it to type.

contractsV5/KyberHintHandler.sol
ayobuenavista: Can't compare of types bytes memory and bytes memory. Added another set of constants which is the keccak256 of the opcode.

contractsV5/KyberHintHandler.sol
ayobuenavista: Using the keccak(opcode) constants instead for comparison.

contractsV5/KyberHintHandler.sol
ayobuenavista: Implemented.

contractsV5/KyberHintHandler.sol
ayobuenavista: Implemented.

contractsV5/KyberHintHandler.sol
ayobuenavista: Changed.

contractsV5/KyberHintHandler.sol
```diff
@@ -20,126 +25,82 @@ contract KyberHintParser is IKyberHint, Utils {
...
     {
         TradeHint memory tradeHint;
         uint indexToContinueFrom;
 
         decodeOperation(hint, tradeHint, indexToContinueFrom, false);
 
-        ethToTokenType = tradeHint.ethToTokenReserves.tradeType;
-        ethToTokenAddresses = tradeHint.ethToTokenReserves.addresses;        
-        ethToTokenSplits = tradeHint.ethToTokenReserves.splitValuesBps;
+        require(!tradeHint.ethToTokenReserves.invalidHint, "Invalid hint");
```
ilanDoron: seems to me if we can't return failing index.
at least we could revert with a clearer reason and some index. and not revert in one centralized place.
lets see how failing revert msgs can be as helpful as possible. also for our own testing.

contractsV5/KyberHintHandler.sol
```diff
@@ -223,85 +184,127 @@ contract KyberHintParser is IKyberHint, Utils {
...
         } else {
-            reserves = tradeHint.ethToTokenReserves;
+            reservesHint = tradeHint.ethToTokenReserves;
         }
 
         indexToContinueFrom += 1;
+        reservesHint.invalidHint = false;
 
-        if (opcodeHash == keccak256(END_OPCODE)) {
+        if (opcodeHash == END_KECCAK) {
```
ilanDoron: seems to me more readable with consistent naming.

so can change opcodeHash to opcodeKeccak
or change END_KECCAK to END_HASH

contractsV5/KyberHintHandler.sol
```diff
@@ -223,85 +184,127 @@ contract KyberHintParser is IKyberHint, Utils {
...
-        } else if (opcodeHash == keccak256(parseOpcode(TradeType.Split))) {
-            reserves.tradeType = TradeType.Split;
-            (reserves.addresses, reserves.splitValuesBps, indexToContinueFrom) = decodeReservesFromHint(true, hint, indexToContinueFrom);
+        } else if (opcodeHash == SPLIT_TRADE_KECCAK) {
+            reservesHint.tradeType = TradeType.Split;
+            (indexToContinueFrom) = decodeReservesFromHint(true, hint, reservesHint, indexToContinueFrom);
             decodeOperation(hint, tradeHint, indexToContinueFrom, isTokenToEth);
-        } else {
-            revert("Invalid hint opcode");
+        } else {
```
ilanDoron: added un required space here.

contractsV5/KyberHintHandler.sol
```diff
@@ -223,85 +184,127 @@ contract KyberHintParser is IKyberHint, Utils {
...
-            reserves.tradeType = TradeType.Split;
-            (reserves.addresses, reserves.splitValuesBps, indexToContinueFrom) = decodeReservesFromHint(true, hint, indexToContinueFrom);
+        } else if (opcodeHash == SPLIT_TRADE_KECCAK) {
+            reservesHint.tradeType = TradeType.Split;
+            (indexToContinueFrom) = decodeReservesFromHint(true, hint, reservesHint, indexToContinueFrom);
             decodeOperation(hint, tradeHint, indexToContinueFrom, isTokenToEth);
-        } else {
-            revert("Invalid hint opcode");
+        } else {            
+            reservesHint.invalidHint = true;
```
ilanDoron: seems better to revert from 2 reasons.
1. gives user a better idea why it reverted (if its on chain call)
2. on chain it saves user gas to revert as early as possible.
when reverting also add current index to message.

contractsV5/KyberHintHandler.sol
```diff
@@ -223,85 +184,127 @@ contract KyberHintParser is IKyberHint, Utils {
...
                splitValuesBps[i] = uint(hint.toUint16(indexToContinueFrom));
                bpsSoFar += splitValuesBps[i];
                indexToContinueFrom += 2;
            }
        }
+
        require((bpsSoFar == BPS) || (bpsSoFar == 0), "bps do not sum to 10000");
-       return (addresses, splitValuesBps, indexToContinueFrom);
+
+       reservesHint.addresses = addresses;
```
ilanDoron: why not set directly to memory.
why do we need an extra parameter?

contractsV5/KyberTradeLogic.sol
```diff
@@ -473,11 +469,101 @@ contract KyberTradeLogic is KyberHintParser, IKyberTradeLogic, PermissionGroups
         return (reserveArr[bestReserve.index], rates[bestReserve.index], isFeePaying);
     }
 
-    function convertReserveIdToAddresses(bytes8 reserveId)
+    function parseEthToTokenHint(bytes calldata hint)
+        external
```
ilanDoron: I don't follow why is this function in trade logic.
it can't be in hintHandler?
only function it need from trade logic is : convertAddressToReserveId
and once trade logic implements this function, it can use it and parse hint in hint handler

or maybe I am missing something.
lets discuss.

can only function that should be in trade logic

contractsV5/KyberTradeLogic.sol
```diff
@@ -473,11 +469,101 @@ contract KyberTradeLogic is KyberHintParser, IKyberTradeLogic, PermissionGroups
...
+            TradeType ethToTokenType,
+            bytes8[] memory ethToTokenReserveIds,
+            uint[] memory ethToTokenSplits
+        )
+    {
+        IKyberReserve[] memory ethToTokenAddresses;
+
+        (ethToTokenType, ethToTokenAddresses, ethToTokenSplits) = parseHintE2T(hint);
+
+        bytes8[] memory reserveIds = new bytes8[](ethToTokenAddresses.length);
```
ilanDoron: why declare an extra paramter and later asign to an unused parameter.
it makes code un necessary long.
longer code is less readable.
also might be more gas to allocate extra variable.
unless compiler is smart enough. which I am not sure.

contractsV5/KyberTradeLogic.sol
```diff
@@ -473,11 +469,101 @@ contract KyberTradeLogic is KyberHintParser, IKyberTradeLogic, PermissionGroups
...
+            TradeType tokenToEthType,
+            bytes8[] memory tokenToEthReserveIds,
+            uint[] memory tokenToEthSplits
+        )
+    {
+        IKyberReserve[] memory tokenToEthAddresses;
+
+        (tokenToEthType, tokenToEthAddresses, tokenToEthSplits) = parseHintT2E(hint);
+
+        bytes8[] memory reserveIds = new bytes8[](tokenToEthAddresses.length);
```
ilanDoron: please remove extra parameter.

contractsV5/KyberTradeLogic.sol
```diff
@@ -473,11 +469,101 @@ contract KyberTradeLogic is KyberHintParser, IKyberTradeLogic, PermissionGroups
...
+        (
+            tokenToEthType,
+            tokenToEthAddresses,
+            tokenToEthSplits,
+            ethToTokenType,
+            ethToTokenAddresses,
+            ethToTokenSplits
+        ) = parseHintT2T(hint);
+
+        bytes8[] memory reserveIdsT2E = new bytes8[](tokenToEthAddresses.length);
```
ilanDoron: please remove

contractsV5/KyberHintHandler.sol
```diff
@@ -223,85 +184,127 @@ contract KyberHintParser is IKyberHint, Utils {
...
            indexToContinueFrom += RESERVE_ID_LENGTH;
+
            if (isTokenSplit) {
                splitValuesBps[i] = uint(hint.toUint16(indexToContinueFrom));
                bpsSoFar += splitValuesBps[i];
                indexToContinueFrom += 2;
            }
        }
+
        require((bpsSoFar == BPS) || (bpsSoFar == 0), "bps do not sum to 10000");
```
ilanDoron: shorter message.
maybe BPS > 10000
and add indexToContinueFrom
to get more helpful message.

contractsV5/mock/MockHintHandler.sol
```diff
@@ -0,0 +1,113 @@
...
+
+    mapping(address=>bytes8) public reserveAddressToId;
+    mapping(bytes8=>address[]) public reserveIdToAddresses;
+
+    function addReserve(address reserve, bytes8 reserveId) public {
+        reserveIdToAddresses[reserveId].push(reserve);
+        reserveAddressToId[reserve] = reserveId;
+    }
+
+    function parseEthToTokenHint(bytes calldata hint)
```
ilanDoron: seems these functions should be in hintHandler.
and here only need to implement
IDtoAddress
AddressToId

contractsV5/IKyberHint.sol
```diff
@@ -13,20 +13,18 @@ interface IKyberHint {
...
 
     function parseTokenToEthHint(bytes calldata hint)
         external
         view
         returns(
-            TradeType tradeType,
-            bytes8[] memory reserveIds,
-            uint[] memory splits,
-            uint failingIndex
+            TradeType tokenToEthType,
```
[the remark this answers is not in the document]
ayobuenavista: Makes sense. Shortened the names for these vars.

contractsV5/KyberHintHandler.sol
ayobuenavista: Changed to opcodeHash to opecodeKeccak for readability.

contractsV5/KyberHintHandler.sol
ayobuenavista: Ah I missed this. Fixed it remove the extra param.

contractsV5/KyberTradeLogic.sol
ayobuenavista: Ah yes. Was thinking that the external functions needed to be in KyberTradeLogic based on our last discussion. Anyway it is in KyberHintHandler now.

contractsV5/KyberTradeLogic.sol
ayobuenavista: Done

contractsV5/KyberTradeLogic.sol
ayobuenavista: Done

contractsV5/KyberHintHandler.sol
ayobuenavista: Done

contractsV5/mock/MockHintHandler.sol
ayobuenavista: Moved

contractsV5/KyberHintHandler.sol
ayobuenavista: Haven't really worked on the failingIndex. Problem is revert/require reason must be a string and can't insert variables in the message.

contractsV5/KyberHintHandler.sol
ayobuenavista: Fixed

contractsV5/KyberHintHandler.sol
ayobuenavista: Will think more about failingIndex. Variables cannot be added to revert reason as mentioned above, and only static string.

contractsV5/KyberTradeLogic.sol
ayobuenavista: Fixed

contractsV5/KyberHintHandler.sol
ilanDoron: at this stage can change to revert.
invalid opcode string (issued here) is better then invalid hint
later can advance with failing index.

contractsV5/KyberHintHandler.sol
```diff
@@ -0,0 +1,352 @@
...
+
+        for (uint i = 0; i < t2eAddresses.length; i++) {
+            t2eReserveIds[i] = convertAddressToReserveId(address(t2eAddresses[i]));
+        }
+        for (uint i = 0; i < e2tAddresses.length; i++) {
+            e2tReserveIds[i] = convertAddressToReserveId(address(e2tAddresses[i]));
+        }
+    }
+
+    function convertReserveIdToAddress(bytes8 reserveId) internal view returns (address) {}
```
ilanDoron: this should have empty body.
so this contract can't be deployed without implementing the functions.

ayobuenavista: Fixed

contractsV5/KyberHintHandler.sol
ayobuenavista: Changed to revert() and removed invalidHint for now.

PR #619, opened 2020-02-06
Fix KyberTradeLogic

Anyhowclick:


contractsV5/KyberTradeLogic.sol
```diff
@@ -212,49 +211,51 @@ contract KyberTradeLogic is KyberHintParser, IKyberTradeLogic, PermissionGroups
...
 
-            require(failingIndex > 0);
+        //init rates and feePaying arrays if split hint
+        if (tradeData.tokenToEth.tradeType == TradeType.Split) {
+            tradeData.tokenToEth.rates = new uint[](tradeData.tokenToEth.addresses.length);
+            tradeData.tokenToEth.isFeePaying = new bool[](tradeData.tokenToEth.addresses.length);
         }
+
+        if (tradeData.ethToToken.tradeType == TradeType.Split) {
+            tradeData.ethToToken.rates = new uint[](1);
```
ilanDoron: seems to me numbers should be as above.
so if trade type is split.
should be split values and fee paying values same as number of reserves.

contractsV5/KyberTradeLogic.sol
```diff
@@ -212,49 +211,51 @@ contract KyberTradeLogic is KyberHintParser, IKyberTradeLogic, PermissionGroups
...
+                tradeData.ethToToken.splitValuesBps
+            ) = parseHintT2T(hint);
+        }
 
-            require(failingIndex > 0);
+        //init rates and feePaying arrays if split hint
+        if (tradeData.tokenToEth.tradeType == TradeType.Split) {
+            tradeData.tokenToEth.rates = new uint[](tradeData.tokenToEth.addresses.length);
+            tradeData.tokenToEth.isFeePaying = new bool[](tradeData.tokenToEth.addresses.length);
         }
```
ilanDoron: don't we want to init with size 1 for the else option?

PR #620, opened 2020-02-06
add proxy tests, update proxy API

ilanDoron:


contractsV5/IKyberNetworkProxy.sol
```diff
@@ -17,16 +17,13 @@ interface IKyberNetworkProxy {
         uint minConversionRate, address payable platformWallet) external payable returns(uint);
 
     // new APIs
-     function getExpectedRateWithHint(IERC20 src, IERC20 dest, uint srcQty, bytes calldata hint) external view
-        returns (uint expectedRate);
-
-    function getExpectedRateAfterCustomFee(IERC20 src, IERC20 dest, uint srcQty, uint customFeeBps, bytes calldata hint) 
+    function getExpectedRateAfterFee(IERC20 src, IERC20 dest, uint srcQty, uint customFeeBps, bytes calldata hint) 
         external view
         returns (uint expectedRate);
```
Anyhowclick: It's clear enough that this rate returned is after network and custom fees right...?

Anyhowclick: I'll merge the changes first, we can discuss if we need to make it clear what expectedRate is being returned.

PR #626, opened 2020-02-11
[WIP] Add Trade Logic Tests

Anyhowclick:
- Remove overwriting of reserveInstances in networkHelper
- Split optionalAmount to optionalBuy and optionalSell amounts
- Use `BN.gte` instead of `>=` for comparison in helper
- Modify getRatesForToken function to reflect rate with network fee
- Added TradeLogic tests

### DIFF
--- a/test/v5/kyberTradeLogic.js
+++ b/test/v5/kyberTradeLogic.js
@@ -1,436 +1,549 @@
 const TestToken = artifacts.require("Token.sol");
 const MockReserve = artifacts.require("MockReserve.sol");
 const MockDao = artifacts.require("MockDAO.sol");
-const KyberNetwork = artifacts.require("KyberNetwork.sol");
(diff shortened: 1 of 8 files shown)


test/v5/kyberTradeLogic.js
```diff
@@ -1,436 +1,549 @@
...
+            });
+
+            it("T2T, split hint", async() => {
+
+            });
+        })
+    })
+});
+
+async function fetchReservesRatesFromTradeLogic(tradeLogicInstance, reserveInstances, tokenAddress, qty, takerFeeBps, isTokenToEth) {
```
ilanDoron: can't we use the global function in nw helper?

no logic is repeated.

I have one concern, if we find a bug in fetch function in nwHelper. and we fix it.
it will not be fixed here.
so can we combine them?

PR #633, opened 2020-02-16
[WIP] don't pull. reduce data packing

ilanDoron:


ilanDoron:
we need some more research on best data packing.

PR #634, opened 2020-02-17
Clicky/add trade logic tests

Anyhowclick:


test/v5/kyberTradeLogic.js
```diff
@@ -19,8 +19,10 @@ const negligibleRateDiffBps = new BN(5); //0.05%
 const maxDestAmt = new BN(2).pow(new BN(255));
 const minConversionRate = new BN(0);
 
-let takerFeeBps = new BN(20);
-let platformFeeBps = new BN(0);
+let takerFeeArray = [new BN(0), new BN(250), new BN(400)];
+let platformFeeArray = [new BN(0), new BN(250, new BN(400))];
```
ilanDoron: I think its more readable if these 2 arrays above are near the loop.
and not here.

lets think of tests as separate from one another.
its clearer for the reader and runner of tests.

so if I decide to run only a specific test and if fails.
all relevant parameters are near by.

test/v5/kyberTradeLogic.js
```diff
@@ -35,7 +37,14 @@ let platformWallet;
 //////////////
 let reserveInstances = {};
 let reserve;
+let bestReserve;
```
ilanDoron: see if all these new parameters can by local and not global.

test/v5/kyberTradeLogic.js
```diff
@@ -347,17 +371,15 @@ contract('KyberTradeLogic', function(accounts) {
                 };
             });
 
-            beforeEach("select random token", async() => {
-                token = (Math.random() > 0.5) ? srcToken : destToken;
+            beforeEach("use srcToken as token", async() => {
```
ilanDoron: can consider looping tokens here.

test/v5/kyberTradeLogic.js
```diff
@@ -563,10 +579,8 @@ contract('KyberTradeLogic', function(accounts) {
 
             it("T2T, mask in hint (both ways)", async() => {
                 numMaskedReserves = 2;
-                for (let takerFee = 0; takerFee <= 1000; takerFee += 500) {
-                    takerFeeBps = new BN(takerFee);
-                    for (let platformFee = 0; platformFee <= 1000; platformFee += 500) {
-                        platformFeeBps = new BN(platformFee);
+                for (takerFeeBps of takerFeeArray) {
```
ilanDoron: I see its repeating many times
isn't it better to add mask type loop here?

test/v5/kyberTradeLogic.js
```diff
@@ -690,63 +698,513 @@ contract('KyberTradeLogic', function(accounts) {
...
+                        expectedSplitValuesBps = expectedSplitValuesBps.concat(BPS);
+                        expectedFeePaying = expectedFeePaying.concat(false);
+
+                        reserveRates = hintedReserves.reservesE2T.reservesForFetchRate.map(reserve => reserve.rate);
+                        expectedTradeResult = getTradeResult(
+                            ethDecimals, [], [], [],
+                            destDecimals, hintedReserves.reservesE2T.reservesForFetchRate, reserveRates, hintedReserves.reservesE2T.splits,
+                            ethSrcQty, takerFeeBps, platformFeeBps);
+                        
+                        expectedReserves = expectedReserves.concat(hintedReserves.reservesE2T.reservesForFetchRate.map(reserve => reserve.address));
```
ilanDoron: these 3 lines are repeating.
Can't make it a function?

test/v5/kyberTradeLogic.js
```diff
@@ -690,63 +698,513 @@ contract('KyberTradeLogic', function(accounts) {
...
+                        expectedFeePaying = expectedFeePaying.concat(hintedReserves.reservesE2T.reservesForFetchRate.map(reserve => reserve.isFeePaying));
+                        
+                        actualResult = await tradeLogic.calcRatesAndAmounts(srcToken.address, destToken.address, srcQty, fees, hintedReserves.hint);
+                        compareResults(expectedTradeResult, expectedReserves, expectedRates, expectedSplitValuesBps, expectedFeePaying, actualResult);
+                    }   
+                }
             });
 
             it("T2T, mask out hint | no hint", async() => {
+                for (takerFeeBps of takerFeeArray) {
```
ilanDoron: wow all combinations?
see if can reduce number of lines using helper function. so make the code shorter.
or looping combinations

test/v5/kyberTradeLogic.js
Anyhowclick: Looping through the different mask types? The expected result is calculated slightly differently for each

PR #637, opened 2020-02-18
update logic to enable working without DAO, update fee handler

ilanDoron:


contractsV5/KyberNetwork.sol
```diff
@@ -15,6 +15,9 @@ import "./IGasHelper.sol";
 /// @title Kyber Network main contract
 contract KyberNetwork is Withdrawable, Utils, IKyberNetwork, ReentrancyGuard {
 
+    uint  constant PERM_HINT_GET_RATE = 1 << 255; //for backwards compatibility
```
Anyhowclick: We need this for...?

PR #639, opened 2020-02-19
Fix some logics in fee handler

manhlx3006:
Some logics are wrong in fee handler, please see my changes

### DIFF
--- a/contractsV5/FeeHandler.sol
+++ b/contractsV5/FeeHandler.sol
@@ -27,6 +27,7 @@ contract FeeHandler is IFeeHandler, Utils {
 
     mapping(address => uint) public rebatePerWallet;
     mapping(uint => uint) public rewardsPerEpoch;
+    mapping(uint => uint) public rewardsPayedPerEpoch;
     uint public totalValues; // total rebates, total rewards
 
     constructor(
@@ -83,11 +84,13 @@ contract FeeHandler is IFeeHandler, Utils {
         // Decoding BRR data
         (uint rewardInBPS, uint rebateInBPS, uint epoch) = getBRR();
 
-        uint rebateWei = rebateInBPS * msg.value / BPS;
-        uint rewardWei = rewardInBPS * msg.value / BPS;
+        uint fee = msg.value;
+
+        uint rebateWei = rebateInBPS * fee / BPS;
+        uint rewardWei = rewardInBPS * fee / BPS;
         for(uint i = 0; i < eligibleWallets.length; i ++) {
             // Internal accounting for rebates per reserve wallet (rebatePerWallet)
-            rebatePerWallet[eligibleWallets[i]] = rebateWei * rebatePercentages[i] / 100;
+            rebatePerWallet[eligibleWallets[i]] += rebateWei * rebatePercentages[i] / 100;
         }
 
         (uint totalRewards , uint totalRebates) = decodeTotalValues(totalValues);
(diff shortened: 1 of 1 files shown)


contractsV5/FeeHandler.sol
```diff
@@ -83,11 +84,13 @@ contract FeeHandler is IFeeHandler, Utils {
...
-        uint rebateWei = rebateInBPS * msg.value / BPS;
-        uint rewardWei = rewardInBPS * msg.value / BPS;
+        uint fee = msg.value;
+
+        uint rebateWei = rebateInBPS * fee / BPS;
+        uint rewardWei = rewardInBPS * fee / BPS;
         for(uint i = 0; i < eligibleWallets.length; i ++) {
             // Internal accounting for rebates per reserve wallet (rebatePerWallet)
-            rebatePerWallet[eligibleWallets[i]] = rebateWei * rebatePercentages[i] / 100;
+            rebatePerWallet[eligibleWallets[i]] += rebateWei * rebatePercentages[i] / 100;
```
manhlx3006: Why here it is `/ 100`? we don't use BPS or Precision for percentage?

contractsV5/FeeHandler.sol
```diff
@@ -117,9 +120,11 @@ contract FeeHandler is IFeeHandler, Utils {
         // Update total rewards and total rewards per epoch
         (uint totalRewards , uint totalRebates) = decodeTotalValues(totalValues);
         require(totalRewards >= amount, "Amount underflow");
-        
-        rewardsPerEpoch[epoch] = rewardsPerEpoch[epoch] - amount;
```
manhlx3006: Note: If you deduct `rewardsPerEpoch[epoch]` when a staker claims the reward, the next staker will get less reward than the actual.

For example: epoch reward is 100 eth, the first stake has 50%, the second has 50%, so by right each should get 50eth. However, after the first staker claimed the reward, epoch reward's reduced to 50 -> the second staker can only claim 50% * 50 = 25 eth.

I added `rewardsPayedPerEpoch` just for accounting

contractsV5/FeeHandler.sol
```diff
@@ -135,12 +140,15 @@ contract FeeHandler is IFeeHandler, Utils {
 
     // Using rebateWallet instead of reserve so I don't have to store KyberNetwork variable.
     function claimReserveRebate(address rebateWallet) external returns (uint){
+        require(rebatePerWallet[rebateWallet] > 1, "no rebate to claim");
```
manhlx3006: safe check here since you can pass any address as `rebateWallet`

contractsV5/FeeHandler.sol
```diff
@@ -199,6 +211,8 @@ contract FeeHandler is IFeeHandler, Utils {
             ""
         );
 
+        require(expectedRate > 0, "rate is 0");
```
manhlx3006: Safe check for expected rate before call trade from network

contractsV5/FeeHandler.sol
```diff
@@ -220,16 +234,21 @@ contract FeeHandler is IFeeHandler, Utils {
 
     // if no one voted for an epoch (like epoch 0). no one gets reward. so should burn it.
     function shouldBurnEpochReward(uint epoch) public {
-        if (!kyberDAO.shouldBurnRewardForEpoch(epoch)) return;
+        require(address(kyberDAO) != address(0), "kyberDAO addr missing");
```
manhlx3006: Better to revert so sender could read the error message (instead of seeing tx is success but actually nothing happens)

contractsV5/FeeHandler.sol
```diff
@@ -220,16 +234,21 @@ contract FeeHandler is IFeeHandler, Utils {
...
-        (uint totalRewardWei, ) = decodeTotalValues(totalValues);
+        (uint totalRewardWei, uint totalRebateWei) = decodeTotalValues(totalValues);
+
+        require(totalRewardWei >= rewardAmount, "total reward less than epoch reward");
 
-        require(totalRewardWei >= rewardAmount);
-        
         // any reward we subtract from total values will be burnt later.
         totalRewardWei -= rewardAmount;
+        totalValues = encodeTotalValues(totalRewardWei, totalRebateWei);
```
manhlx3006: As totalRewad is updated, totalValues should also be updated

manhlx3006:
In general, I think we should also use SafeMath, and put reentrancy guard for claim reward/rebate function

PR #646, opened 2020-02-21
Clicky/katalyst deployer

Anyhowclick:
- Deployer script for TradeLogic, Network, Proxy, FeeHandler contracts
- Minor changes to KN.sol and KNP.sol
- Use latest solc version (not really needed)
- Made compileContracts.js to work for both v4 and v5 contracts
- Modified APR deployer script too


package.json
```diff
@@ -28,7 +28,7 @@
     "rlp": "2.2.4",
     "secp256k1": "3.8.0",
     "sha3": "2.1.1",
-    "solc": "0.5.11",
+    "solc": "0.6.3",
```
ilanDoron: why 0.6.3?

web3deployment/liquidityReserve_input.json
```diff
@@ -1,25 +1,29 @@
 {
-  "tokens": {
-    "KNC": {
+  "token": {
       "name": "KyberNetwork",
+      "symbol": "KNC",
       "decimals": 18,
-      "address": "0x4E470dc7321E84CA96FcAEDD0C8aBCebbAEB68C6"
-    }
+      "address": "0xa67bba948bd8372bc9e60b1acfafb8e2db3cc57c"
```
ilanDoron: is this for ropsten?
we should organize this folder.
All ropsten stuff should move to ropsten folder.
we don't want someone using this thinking its main net KNC
and this folder is not very cluttered.

package.json
Anyhowclick: as discussed, will fix if 0.5.11 suffices

web3deployment/liquidityReserve_input.json
Anyhowclick: Changing it back to the ropsten KNC address. For deployment, project teams will be using replacing the token field with their own token.

Agree that we should organize the folder

Anyhowclick: I leave it as is first, we can re-org the folder in another PR =)

web3deployment/KatalystDeploy/compileContractsV5.js
```diff
@@ -1,94 +0,0 @@
...
-    // console.log(config)
-    let output = compileSources(config);
-    errorHandling(output);
-    // console.log(output.contracts['GasHelper.sol']);
-    // console.log(output);
-
-    return output;
-}
-
-main();
```
ilanDoron: 1 more thing I added here which I found helpful.
if contract is called with no script it just compiles. and better if it also output result.
so if anyone updates the compile contracts file. he can run it without needing any script for checking the update is working as expected.
So I suggest each compile contracts file will have 1 export function
and main function that calls the compile function and prints output.errors.

web3deployment/compileContracts.js
```diff
@@ -1,55 +1,161 @@
...
+        setTimeout(resolve,ms)
+    })
+}
+
+module.exports.compileContracts = compileContracts;
+
+async function compileContracts(versionNum) {
+    let solcVersionNum;
+    let sourceFiles;
+    if (versionNum == "v4") {
```
ilanDoron: its great u could work this out!!

web3deployment/KatalystDeploy/compileContractsV5.js
```diff
@@ -1,94 +0,0 @@
-const fs = require('fs-extra');
```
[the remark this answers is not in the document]
Anyhowclick: compileContracts.js allows for either v4 or v5. I tried modifying to allow both, but the async await messes with the compiler variable. What I mean is that if we compile v5, then v4, the v4 compiler doesn't download and load up in time for compilation, so I get an error.....

I've not found a good fix for it yet... Nice to have, medium priority at this time? All u need to do is call the the function twice! =)

web3deployment/KatalystDeploy/katalystNetworkDeployer.js
```diff
@@ -0,0 +1,526 @@
+#!/usr/bin/env node
+
```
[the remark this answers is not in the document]
Anyhowclick: Ah.. so 
Stage 1: Deploy contracts, save as json.
Stage 2: Setup contracts

The functions are modularized enough to allow this =)

PR #653, opened 2020-02-24
fix revert messages

ilanDoron:


contractsV5/KyberNetwork.sol
```diff
@@ -785,9 +785,9 @@ contract KyberNetwork is Withdrawable, Utils, IKyberNetwork, ReentrancyGuard {
         require(src != dest, "src = dest");
 
         if (src == ETH_TOKEN_ADDRESS) {
-            require(msg.value == srcAmount, "ETH low");
+            require(msg.value == srcAmount, "eth qty !=");
```
manhlx3006: Last time I remember the special characters are not showing nicely on etherscan

manhlx3006: Here are message from etherscan:
`This is a simple error message ! @ # $ % ^ &amp; * ( ) &lt; &gt;`
here is the original one
`This is a simple error message ! @ # $ % ^ & * ( ) < >`

ilanDoron:
will do in another PR

PR #654, opened 2020-02-25
Added tests for BytesLib.sol

ayobuenavista:
Use `truffle test <file>` to run the test.
e.g.
`truffle test test/v5/solidity/TestBytesLib1.sol`
`truffle test test/v5/solidity/TestBytesLib2.sol`

Prerequisites for this test to run is Migrations.sol and AssertBytes.sol.


ayobuenavista:
Closing due to in favor of using abi.[encode/decode] method.

PR #656, opened 2020-02-25
Ilan/network and a handler tests

ilanDoron:


ilanDoron:
will handle in another PR

PR #673, opened 2020-03-02
Adding failedIndex for KyberHintHandler.sol

ayobuenavista:
Modified KyberHintHandler.sol to add accurate failedIndex for incorrect hints.


contractsV5/KyberHintHandler.sol
```diff
@@ -169,89 +183,86 @@ contract KyberHintHandler is IKyberHint, Utils {
...
-            reservesHint.tradeType = TradeType.Split;
-            (indexToContinueFrom) = decodeReservesFromHint(true, hint, reservesHint, indexToContinueFrom);
-            decodeOperation(hint, tradeHint, indexToContinueFrom, isTokenToEth);
+            decodeReservesFromHint(true, hint, TradeType.Split, tradeHint, isTokenToEth);
+            decodeOperation(hint, tradeHint, isTokenToEth);
         } else {
-            revert("Invalid hint opcode");
-        }
+            tradeHint.hintError = true;
+        }
```
ilanDoron: please remove extra spaces

contractsV5/KyberTradeLogic.sol
```diff
@@ -282,9 +285,12 @@ contract KyberTradeLogic is KyberHintHandler, IKyberTradeLogic, PermissionGroups
                 tData.tokenToEth.splitValuesBps,
                 tData.ethToToken.tradeType,
                 tData.ethToToken.addresses,
-                tData.ethToToken.splitValuesBps
+                tData.ethToToken.splitValuesBps,
+                failedIndex
             ) = parseHintT2T(hint);
         }
+
+        require(failedIndex > 0, "Hint error");
```
ilanDoron: notice for get rate should return 0 and not revert.
consult Desmond how to know which is the calling function.

ilanDoron: should probably return some error code from this function.
and calling function should return 0 rate.
trade knows to revert on 0 rate.
so this is how it should join the trade flow.

PR #697, opened 2020-03-05
Added test for gas helper

ayobuenavista:


test/sol5/kyberNetwork.js
```diff
@@ -384,10 +386,13 @@ contract('KyberNetwork', function(accounts) {
             rateHelper = await RateHelper.new(admin);
             await rateHelper.setContracts(matchingEngine.address, DAO.address, {from: admin});
 
+            // init gas helper
```
ilanDoron: please add a comment how is this tested.

test/sol5/kyberNetwork.js
```diff
@@ -876,6 +881,32 @@ contract('KyberNetwork', function(accounts) {
...
+            after("unlist and remove reserve", async() => {
+                await nwHelper.removeReservesFromNetwork(network, reserveInstances, tokens, operator);
+                reserveInstances = {};
+            });
+
+            it("test that gas helper can't revert trade even if it reverts", async() => {
+                platformFeeBps = new BN(50);
+                hint = await nwHelper.getHint(rateHelper, matchingEngine, reserveInstances, hintType, undefined, ethAddress, destToken.address, ethSrcQty);
+
+                await network.tradeWithHintAndFee(network.address, ethAddress, ethSrcQty, destToken.address, taker,
```
ilanDoron: can you somehow show it reverted internally?
maybe print tx reciept and see if this data exists there.
not sure about it

test/sol5/kyberNetwork.js
```diff
@@ -876,6 +881,32 @@ contract('KyberNetwork', function(accounts) {
...
+
+            after("unlist and remove reserve", async() => {
+                await nwHelper.removeReservesFromNetwork(network, reserveInstances, tokens, operator);
+                reserveInstances = {};
+            });
+
+            it("test that gas helper can't revert trade even if it reverts", async() => {
+                platformFeeBps = new BN(50);
+                hint = await nwHelper.getHint(rateHelper, matchingEngine, reserveInstances, hintType, undefined, ethAddress, destToken.address, ethSrcQty);
+
```
ilanDoron: please add a comment.
if any wallet used that is not platform wallet. gas helper will revert. see mockGasHelper code.

test/sol5/kyberNetwork.js
```diff
@@ -384,10 +386,13 @@ contract('KyberNetwork', function(accounts) {
...
             await rateHelper.setContracts(matchingEngine.address, DAO.address, {from: admin});
 
+            // init gas helper
+            gasHelperAdd = await MockGasHelper.new(platformWallet);
+
             //setup network
             await network.addOperator(operator, {from: admin});
             await network.addKyberProxy(networkProxy, {from: admin});
-            await network.setContracts(feeHandler.address, matchingEngine.address, zeroAddress, {from: admin});
+            await network.setContracts(feeHandler.address, matchingEngine.address, gasHelperAdd.address, {from: admin});
```
ilanDoron: if this is the same network used in other tests.
better remove this  gas helper after the test.

test/sol5/kyberNetwork.js
ayobuenavista: It's hard to see in tx receipt, but I was able to check the condition that will makes it revert. Also, checked by changing `address(gasHelper).call()...` to `gasHelper.freeGas()...` to confirm that it reverts.

test/sol5/kyberNetwork.js
ayobuenavista: Added comment.

test/sol5/kyberNetwork.js
ayobuenavista: This should be okay since this change is only used within a describe. The other tests initialize another network.

PR #705, opened 2020-03-06
Adding reserveIDs as return to MatchingEngine from HintHandler

ayobuenavista:


ayobuenavista:
Stack too deep error still exists..

contractsSol5/KyberMatchingEngine.sol
```diff
@@ -437,7 +467,7 @@ contract KyberMatchingEngine is KyberHintHandler, IKyberMatchingEngine, Withdraw
             rates[i] = tData.tokenToEth.rates[i];
             splitValuesBps[i] = tData.tokenToEth.splitValuesBps[i];
             isFeePaying[i] = tData.tokenToEth.isFeePaying[i];
-            ids[i] = convertAddressToReserveId(address(reserveAddresses[i]));
```
ayobuenavista: I think main issue is that we want to get rid of this. But looking back at it now, it doesn't seem too bad since we the loop already exists anyway to populate the other params.

contractsSol5/KyberMatchingEngine.sol
```diff
@@ -446,7 +476,7 @@ contract KyberMatchingEngine is KyberHintHandler, IKyberMatchingEngine, Withdraw
             rates[i] = tData.ethToToken.rates[i - tokenToEthNumReserves];
             splitValuesBps[i] = tData.ethToToken.splitValuesBps[i - tokenToEthNumReserves];
             isFeePaying[i] = tData.ethToToken.isFeePaying[i - tokenToEthNumReserves];
-            ids[i] = convertAddressToReserveId(address(reserveAddresses[i]));
```
ayobuenavista: Same as above.

contractsSol5/KyberHintHandler.sol
```diff
@@ -336,32 +325,18 @@ contract KyberHintHandler is IKyberHint, Utils4 {
...
+            ethToTokenReserveIds,
             ethToTokenSplits,
             failedIndex
         ) = parseHintT2T(hint);
-
-        if (failedIndex == 0) {
-            tokenToEthReserveIds = new bytes8[](tokenToEthAddresses.length);
-            ethToTokenReserveIds = new bytes8[](ethToTokenAddresses.length);
-
-            for (uint i = 0; i < tokenToEthAddresses.length; i++) {
```
ayobuenavista: So we wanted to remove this to avoid more storage reads and additional loop, and to also remove one of the virtual functions. But this is part of an external function anyway, so we didn't care if there was additional logic.

contractsSol5/KyberMatchingEngine.sol
ilanDoron: my issue is the extra storage read not the loop
if we handle 8 reserves
its extra 8 * 800 gas
while we had this value to start with. so makes no sense reading it here

contractsSol5/KyberMatchingEngine.sol
ilanDoron: same reply

ilanDoron:
we can consider this change for next version.
for now locked the code for audit.

PR #708, opened 2020-03-08
Ilan/update reserve type setting

ilanDoron:


contractsSol5/KyberNetwork.sol
```diff
@@ -755,7 +774,7 @@ contract KyberNetwork is Withdrawable2, Utils4, IKyberNetwork, ReentrancyGuard {
         require(tData.rateOnlyNetworkFee >= tData.input.minConversionRate, "rate < minConvRate");
 
         if (gasHelper != IGasHelper(0)) {
-            address(gasHelper).call(
+            (bool success, ) = address(gasHelper).call(
```
Anyhowclick: Is it possible to not have the `(bool success)` and just do the function call. Since we don't care if it reverts or not.

contractsSol5/KyberMatchingEngine.sol
```diff
@@ -5,16 +5,29 @@ import "./IKyberMatchingEngine.sol";
 import "./IKyberNetwork.sol";
 import "./KyberHintHandler.sol";
 
-
+/*
+*   @title Kyber matching engine contract
+*   Recieves call from KyberNetwork for:
```
Anyhowclick: Receives

PR #717, opened 2020-03-09
move gas helper to mock folder

ilanDoron:


ilanDoron:
will do in another PR

PR #741, opened 2020-03-14
add travis ci for sol5 test

ductm54:


.travis/test-part.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+# -*- firestarter: "shfmt -i 4 -ci -w %p" -*-
+
+set -euxo pipefail
+
+readonly test_part=${TEST_PART:-}
+
+case "$test_part" in
+Sol5)
+    npx buidler test --no-compile --config ./buidlerCoverageSol5.js
```
ilanDoron: why use coverage here?

can use "buildler.config.js" ??

buidlerCoverageSol4.js
```diff
@@ -7,6 +7,7 @@ module.exports = {
 
   paths: {
     sources: "./contracts",
-    artifacts: ".coverageV4Artifacts"
+    artifacts: ".coverageV4Artifacts",
```
ilanDoron: not sure why use coverage file?

.travis/test-part.sh
ductm54: i just copy this one from tst.sh, will fix

buidlerCoverageSol4.js
ductm54: should not change this file. will remove

.travis/test-part.sh
ductm54: just check again use buidler.config.js will run every test. on this we only want to run sol5 test

.travis/test-part.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+# -*- firestarter: "shfmt -i 4 -ci -w %p" -*-
+
```
ilanDoron: what does this comment mean?

ductm54: i use shfmt to format bash. could remove if needed

PR #742, opened 2020-03-14
Update compileContracts due to folder struct changes, starting block …

Anyhowclick:
…due to deployment

### DIFF
--- a/web3deployment/compileContracts.js
+++ b/web3deployment/compileContracts.js
@@ -183,6 +183,10 @@ async function main() {
     let output = {'contracts': {}, 'sources': {}};
     //TODO: combine outputs
     let v4Output = await compileContracts("sol4");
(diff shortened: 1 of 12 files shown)


web3deployment/compileContracts.js
```diff
@@ -176,4 +185,4 @@ async function main() {
     let v4Output = await compileContracts("sol4");
     let v5Output = await await compileContracts("sol5");
     return [v4Output, v5Output];
-}
\ No newline at end of file
+}
```
ilanDoron: 1 thing missing.

add in file :

main();

so can just run

node compileContracts.js

Anyhowclick:
Done. Should work now

Anyhowclick:
Try again... Hopefully it works now. Problem is that solc has to download the specific version, so it's dependent on ur internet connection

PR #743, opened 2020-03-15
Add full integration tests between all contracts

manhlx3006:


test/sol5/kyberProxyIntegration.js
```diff
@@ -0,0 +1,710 @@
...
+        }
+
+        srcToken = tokens[0];
+        srcDecimals = tokenDecimals[0];
+        srcQty = new BN(10).mul((new BN(10)).pow(new BN(srcDecimals)));
+        destToken = tokens[1];
+        ethSrcQty = precisionUnits;
+
+        // init and setup reserves
+        let result = await nwHelper.setupReserves(network, tokens, 1, 2, 3, 0, accounts, admin, operator);
```
ilanDoron: 3rd type not supported yet. so the '3' has no effect.

manhlx3006: Removed the third type

manhlx3006:
- with no network fee + brr campaigns, the different gas consumption between first and second trade is around 50k+ (check out epoch 0 and 1, where we don't have to conclude any campaigns)
- with both network fee + brr campaigns, the different gas consumption is around 130k.

So concluding 2 campaigns only costs around 70-80k of gas.

PR #747, opened 2020-03-16
Ilan/audit updates

ilanDoron:


ilanDoron:
will PR only after Yaron done with Audit

PR #753, opened 2020-03-17
don't pull - use struct to save storage reads (instead of bit shifting)

ilanDoron:


contractsSol5/KyberFeeHandler.sol
```diff
@@ -303,46 +310,53 @@ contract KyberFeeHandler is IKyberFeeHandler, Utils4 {
         emit RewardsRemovedToBurn(epoch, rewardAmount);
     }
 
-    function encodeBRRData(uint _reward, uint _rebate, uint _epoch, uint _expiryBlock) public pure returns (uint) {
-        return (((((_reward << BITS_PER_PARAM) + _rebate) << BITS_PER_PARAM) + _epoch) << BITS_PER_PARAM) + _expiryBlock;
+    function updateBRRData(uint reward, uint rebate, uint epoch, uint expiryBlock) public {
+        require(reward < 2 ** 32 - 1);
```
Anyhowclick: We agree to have comment for the require statements right? =)

Anyhowclick: Also, swap the epoch and expiryBlock fields? Cos the other functions are in that manner....

Anyhowclick: See my PR made to ur branch

PR #760, opened 2020-03-19
Ilan/remove bit games

ilanDoron:


contractsSol5/KyberFeeHandler.sol
```diff
@@ -303,46 +310,53 @@ contract KyberFeeHandler is IKyberFeeHandler, Utils4 {
...
+        (rewardBps, rebateBps, expiryBlock, epoch) = readBRRData();
 
           // Check current block number
         if (block.number > expiryBlock && kyberDAO != IKyberDAO(0)) {
-            uint burnBPS;
+            uint burnBps;
 
-            (burnBPS, rewardBPS, rebateBPS, epoch, expiryBlock) = kyberDAO.getLatestBRRData();
+            (burnBps, rewardBps, rebateBps, epoch, expiryBlock) = kyberDAO.getLatestBRRData();
+            require(burnBps + rewardBps + rebateBps == BPS, "Bad BRR values");
```
manhlx3006: could bad DAO return BRR that causes overflow here? May be can add a test with bad DAO for the overflow.

Anyhowclick: Yes, it's possible to have overflow.
setMockBRR(10000, 1)
burnBPS = BPS - 10000 - 1 = 2**256 - 1
The check will still pass, but the values are invalid.

Will need more checks here.

PR #766, opened 2020-03-22
Limit network fee value in DAO to less than 50%

manhlx3006:


test/sol5/kyberProxyIntegration.js
```diff
@@ -652,53 +652,258 @@ contract('Proxy + Network + MatchingEngine + FeeHandler + Staking + DAO integrat
         );
     });
 
-    it("test can not record new data for DAO as no trade for more than 1 epoch", async() => {
+    it("test update network fee from DAO with new fee is 0, network updates network fee, feeHandler doesn't update brr data", async() => {
```
ilanDoron: can write
... with fee 0, ...
remove new. 
remove is.

ilanDoron: why fee hanlder doesn't update BRR data (in comment)
 don't follow  this

manhlx3006: in `handleFees` function in FeeHandler, if `feeBRR` is 0, it will return without getting new BRR data from DAO. That's why the BRR data doesn't update

PR #770, opened 2020-03-23
check and test overflow/underflow to Katalyst

ducquangkstn:


test/sol5/kyberMatchingEngine.js
```diff
@@ -2172,110 +2172,22 @@ contract('KyberMatchingEngine', function(accounts) {
                 });
             });
 
-            describe("test fees = BPS", async() => {
```
ilanDoron: why removing this?

contractsSol5/KyberMatchingEngine.sol
```diff
@@ -191,6 +192,10 @@ contract KyberMatchingEngine is KyberHintHandler, IKyberMatchingEngine, Withdraw
         tData.tokenToEth.decimals = srcDecimals;
         tData.ethToToken.decimals = destDecimals;
         tData.networkFeeBps = info[uint(IKyberMatchingEngine.InfoIndex.networkFeeBps)];
+        tData.platformFeeBps = info[uint(IKyberMatchingEngine.InfoIndex.platformFeeBps)];
```
ilanDoron: to avoid long function

please create function initAndValidateTradeData

so we can have TradeData memory tData = initAndValidateTradeData(.,,)
and pass required values.

Anyhowclick: Do we need the checks? Or are we shifting the checks from network contract to this one? Will feel like unnecessary duplicate checks if we do it in both places.

test/sol5/kyberMatchingEngine.js
ducquangkstn: change to update the expected output is revert

contractsSol5/KyberMatchingEngine.sol
ducquangkstn: This is an external view function, it can be called by normal user, so I think it is ok to check it. 
And it will make sense that other math calculation is not overflow with assumption that `networkFeeBps` and `platformFeeBps` less than 10000

contractsSol5/KyberMatchingEngine.sol
```diff
@@ -238,6 +231,21 @@ contract KyberMatchingEngine is KyberHintHandler, IKyberMatchingEngine, Withdraw
         return packResults(tData);
     }
 
+    function initAndValidateTradeData(
+        uint srcDecimals,
+        uint destDecimals,
+        uint[] memory info)
+    internal pure returns(TradeData memory tData) {
+        tData.tokenToEth.decimals = srcDecimals;
```
ilanDoron: please move '{' 
to next line

contractsSol5/KyberMatchingEngine.sol
```diff
@@ -224,14 +222,31 @@ contract KyberMatchingEngine is KyberHintHandler, IKyberMatchingEngine, Withdraw
         //fee deduction
         //ETH -> dest fee deduction has not occured for non-split ETH -> dest trade types
         tData.networkFeeWei = tData.tradeWei * tData.networkFeeBps / BPS * tData.feePayingReservesBps / BPS;
-        tData.platformFeeWei = tData.tradeWei * info[uint(IKyberMatchingEngine.InfoIndex.platformFeeBps)] / BPS;
+        tData.platformFeeWei = tData.tradeWei * tData.platformFeeBps / BPS;
 
+        // this line we can not reach full coverage due to networkFeeWei * 2 + platformFeeWei < BPS and feePayingReservesBps < 2 *BPS
```
Anyhowclick: Can remove this line

contractsSol5/KyberMatchingEngine.sol
```diff
@@ -537,8 +552,9 @@ contract KyberMatchingEngine is KyberHintHandler, IKyberMatchingEngine, Withdraw
                 tData.numFeePayingReserves++;
             }
 
-            //take into account possible additional networkFee
+            // this line we can not reach full coverage due to networkFeeWei * 2 + platformFeeWei < BPS and feePayingReservesBps < 2 *BPS
```
Anyhowclick: You can write a maliciousMatchingEngine with a `initAndValidateTradeData` function that doesn't have the BPS checks. Then you'll reach this line(s)

test/sol5/kyberMatchingEngine.js
ilanDoron: great thnaks

contractsSol5/KyberMatchingEngine.sol
ducquangkstn: change as request

PR #774, opened 2020-03-24
[Don't pull - for comparison] E2T part refactored

Anyhowclick:
- Renamed `storeTradeReserveData` to `storeSingleReserveData`
- Refactor E2T to split and non-split


ilanDoron:
revamp kills this direction

PR #776, opened 2020-03-24
Modified KyberHintHandler to abi.[encode/decode] method

ayobuenavista:


contractsSol5/KyberHintHandler.sol
```diff
@@ -346,22 +177,139 @@ contract KyberHintHandler is IKyberHint, Utils4 {
...
+        } else {
+            if (splits.length != 0) return HintErrors.SplitsZeroError;
+        }
+
+        return HintErrors.NoError;
+    }
+
+    /// @notice Throws error message to user to indicate error on hint
+    /// @param error Error type from HintErrors enum
+    function throwHintError(HintErrors error) internal pure {
```
ilanDoron: please try a bit shorter message
this is still part of matching engine.
so should be readable
but shorter.

contractsSol5/KyberHintHandler.sol
```diff
@@ -346,22 +177,139 @@ contract KyberHintHandler is IKyberHint, Utils4 {
...
+        internal
+        view
+        returns(
+            TradeType t2eType,
+            IKyberReserve[] memory t2eAddresses,
+            uint[] memory t2eSplits,
+            TradeType e2tType,
+            IKyberReserve[] memory e2tAddresses,
+            uint[] memory e2tSplits,
+            HintErrors verified
```
ilanDoron: valid seems to me a bit more suitable.
not sure though

ayobuenavista: Hmmm okay. Using Valid is fine too.

contractsSol5/KyberHintHandler.sol
ayobuenavista: Okay, will try to shorten messages.

ayobuenavista:
Don't merge yet. Adding some missing tests for T2T for duplicate reserveIds. Also fixing some formatting on hintHandler test. :D

PR #781, opened 2020-03-25
more network test

ductm54:


contractsSol5/mock/MockNetwork.sol
```diff
@@ -39,6 +39,10 @@ contract MockNetwork is KyberNetwork {
         (_networkFeeBps, _expiryBlock) = readNetworkFeeData();
     }
 
+    function mockGetNetworkFee() public view returns(uint networkFeeBps) {
```
ilanDoron: instead of mock function
can use this API

getAndUpdateNetworkFee

ductm54: need to test a line in this function mock. if we use getAndUpdateNetworkFee that line could not be reached

PR #782, opened 2020-03-25
Add duplicate ID check error, hintHandler test revamp

Anyhowclick:


contractsSol5/KyberHintHandler.sol
```diff
@@ -256,8 +256,13 @@ contract KyberHintHandler is IKyberHint, Utils4 {
             if (reserveIds.length != splits.length) return false;
 
             uint bpsSoFar;
+            bytes8[] memory checkDuplicateIds = new bytes8[](reserveIds.length);
             for (uint i = 0; i < splits.length; i++) {
                 bpsSoFar += splits[i];
+                for (uint j = 0; j < checkDuplicateIds.length; j++) {
```
ilanDoron: nice
Now this function returns error codes
so should update this.

test/sol5/kyberMatchingEngine.js
```diff
@@ -1663,6 +1663,56 @@ contract('KyberMatchingEngine', function(accounts) {
                 );
             });
 
+            describe(`should return zero rates and amounts if duplicate reserve ID is used for split`, async() => {
```
ilanDoron: should add hint handler test as well.

contractsSol5/KyberHintHandler.sol
Anyhowclick: updated

test/sol5/kyberMatchingEngine.js
Anyhowclick: added

Anyhowclick:
- Use expectRevert instead of try-catch
- Move some "it" tests to before block so that .only works

test/sol5/kyberMatchingEngine.js
Anyhowclick: Done. Also, helped to "re-write" the tests, cos I started fixing 1 small thing that led to fixing another small thing, that led to fixing a big thing.

Just like this: https://www.youtube.com/watch?v=AbSehcT19u0

PR #783, opened 2020-03-25
refactor increase block function

ducquangkstn:


package.json
```diff
@@ -39,7 +39,7 @@
   "devDependencies": {
     "@nomiclabs/buidler": "1.1.2",
     "@nomiclabs/buidler-web3": "1.1.2",
-    "@openzeppelin/test-helpers": "0.5.4",
+    "@openzeppelin/test-helpers": "^0.5.5",
```
ilanDoron: please set as fixed version.
remove ^

So we have control when we change our work environment.

test/sol5/kyberDao.js
```diff
@@ -146,17 +146,15 @@ contract('KyberDAO', function(accounts) {
         await deployContracts(20, currentBlock + 20, 8);
         await setupSimpleStakingData();
 
-        currentBlock = await Helper.getCurrentBlock();
-        await Helper.increaseBlockNumberBySendingEther(accounts[0], accounts[0], startBlock - currentBlock);
+        await time.advanceBlockTo(startBlock);
```
ilanDoron: can;t we call the helper function with this parameter?

test/sol5/kyberDao.js
```diff
@@ -9,7 +9,7 @@ const Helper = require("../helper.js");
 const BN = web3.utils.BN;
 
 const { precisionUnits, zeroAddress } = require("../helper.js");
-const { expectEvent, expectRevert } = require('@openzeppelin/test-helpers');
+const { expectEvent, expectRevert, time } = require('@openzeppelin/test-helpers');
```
ilanDoron: nice if time only in helper
so if later we have issues, easy to change it back.
so all block increase or advance to can be done using our helper.

now its inconsistent.

package.json
ducquangkstn: change as request

test/sol5/kyberDao.js
ducquangkstn: added a function named `increaseBlockNumberTo` which calls `time.advanceBlockTo`

test/sol5/kyberDao.js
ducquangkstn: change as request

PR #827, opened 2020-04-07
Duc/change public to internal

ductm54:


contractsSol5/KyberNetwork.sol
```diff
@@ -65,6 +65,27 @@ contract KyberNetwork is WithdrawableNoModifiers, Utils4, IKyberNetwork, Reentra
         emit EtherReceival(msg.sender, msg.value);
     }
 
+
```
ilanDoron, line 68: spacing between functions should be 1 line.

function is not in a good place.
more important functions should be on top of file.
can also see solidity coding conventions.
A good place in network would be near getNetworkData

contractsSol5/KyberNetwork.sol
```diff
@@ -65,6 +65,27 @@ contract KyberNetwork is WithdrawableNoModifiers, Utils4, IKyberNetwork, Reentra
...
+            IKyberStorage memory storageAddress,
+            IKyberNetworkProxy[] memory proxyAddresses
+        )
+    {
+        return (
+            feeHandler,
+            kyberDAO,
+            matchingEngine,
+            kyberStorage,
+            kyberStorage.getKyberProxies()
```
ilanDoron, line 85: please add gasHelper address

PR #862, opened 2020-04-15
Prettified bridges subfolder

ayobuenavista:


ayobuenavista:
Closing due to bridges already deployed in mainnet and we want the code to be consistent with the deployed ones.

PR #863, opened 2020-04-15
remove network revert msg

ducquangkstn:


test/sol5/kyberNetwork.js
```diff
@@ -173,9 +173,8 @@ contract('KyberNetwork', function(accounts) {
         it("test can't add same proxy twice", async() => {
             await tempNetwork.addKyberProxy(proxy1, {from: admin});
 
-            await expectRevert(
-                tempNetwork.addKyberProxy(proxy1, {from: admin}),
-                "proxy exists"
+            await expectRevert.unspecified(
```
ilanDoron, line 176: if we have valid kyberStorage

We still expect same message. no?

test/sol5/kyberNetwork.js
```diff
@@ -173,14 +173,13 @@ contract('KyberNetwork', function(accounts) {
...
+                tempNetwork.addKyberProxy(proxy1, {from: admin})
             );
         });
 
         it("test can't add proxy zero address", async() => {
-            await expectRevert(
+            await expectRevert.unspecified(
```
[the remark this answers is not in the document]
Anyhowclick: Not all tests, only those that might fail with "invalid opcode", ie. hintHandler tests. Think can leave this one.

test/sol5/kyberNetwork.js
ducquangkstn: `kyberStorage` don't have `mapping(address => bool) internal kyberProxyContracts` so it will fail at` kyberNetwork`
would you like to add `kyberProxyContracts` to `kyberStorage`

ilanDoron: Ok
got it.

PR #866, opened 2020-04-15
Prettified utils & wrappers subfolder

ayobuenavista:


contractsSol5/utils/BytesLib.sol
```diff
@@ -10,10 +10,7 @@ pragma solidity 0.5.11;
 
 
 library BytesLib {
-    function concat(
```
ilanDoron, line 13: can deleted this file...

contractsSol5/utils/zeppelin/Address.sol
```diff
@@ -1,5 +1,6 @@
 pragma solidity 0.5.11;
 
+
```
ilanDoron, line 3: better not to touch zeppellin files.

if we ever want to compare code and see its same...
or pull changes from Zeppelin.

so please avoid updating zepellin.

PR #872, opened 2020-04-16
Simplify withdrawal logics to allow staker withdraw

manhlx3006:
Current with withdraw function has complex logics updating data in Staking and Dao contract
We want to allow staker withdraws their stake even if something is wrong with DAO or updating data in Staking, only need to check if staker's latest stake >= withdrawal amount


contractsSol5/Dao/KyberStaking.sol
```diff
@@ -154,6 +153,9 @@ contract KyberStaking is IKyberStaking, EpochUtils, ReentrancyGuard {
     }
 
     event Withdraw(uint curEpoch, address staker, uint amount);
+    // event is fired if something is wrong with withdrawal
+    // even though the withdrawal is still success
```
ilanDoron, line 157: typo "is still success" -->
will still succeed.
or
is still successful

manhlx3006: Fixed in latest commit

PR #905, opened 2020-04-23
Sol5 -> Sol6

Anyhowclick:
- Not sure how to fix travis, need someone's help. Probably Duc
- ./coverage.sh doesn't work at the moment. Some functions fail to be overridden properly by coverage I think...
- ./test/sol4/kyberProxyV1.js tests are broken... need someone to help fix as well


.solcover.js
```diff
@@ -2,23 +2,23 @@ const fs = require('fs');
 const path = require('path');
 const mv = require('mv');
 
-const v4ArtifactsPath = path.join(__dirname, '.coverageV4Artifacts');
+const oldArtifactsPath = path.join(__dirname, '.coverageArtifacts');
```
ilanDoron, line 5: I don't follow what does old mean here?
can it be named previous?
or is it old?

buidlerConfigSol4.js
```diff
@@ -1,11 +1,10 @@
 module.exports = {
-  defaultNetwork: "buidlerevm",
```
ilanDoron, line 2: this changes?

contracts/sol6/Dao/IEpochUtils.sol
```diff
@@ -0,0 +1,7 @@
+pragma solidity 0.6.6;
+
+interface IEpochUtils {
+    function epochPeriodInSeconds() external view returns (uint256);
+
+    function firstEpochStartTimestamp() external view returns (uint256);
+}
```
ilanDoron, line 7: last line.

contracts/sol6/Dao/KyberFeeHandler.sol
```diff
@@ -172,10 +171,12 @@ contract KyberFeeHandler is IKyberFeeHandler, Utils4, BurnConfigPermission {
         _;
     }
 
-    function() external payable {
+    receive() external payable {
         emit EthReceived(msg.value);
     }
 
+    fallback() external {}
```
ilanDoron, line 178: do we need this one?

contracts/sol6/IKyberDAO.sol
```diff
@@ -1,7 +1,9 @@
-pragma solidity 0.5.11;
+pragma solidity 0.6.6;
 
+import "./Dao/IEpochUtils.sol";
 
-interface IKyberDAO {
+
+interface IKyberDAO is IEpochUtils {
```
ilanDoron, line 6: so now can inherit interface.
its new to me.

contracts/sol6/KyberHintHandler.sol
```diff
@@ -280,7 +300,7 @@ contract KyberHintHandler is IKyberHint, Utils4 {
         if (error == HintErrors.SplitsNotEmptyError) revert("splits must be empty");
     }
 
-    function convertReserveIdToAddress(bytes32 reserveId) internal view returns (address);
+    function convertReserveIdToAddress(bytes32 reserveId) internal view virtual returns (address);
```
ilanDoron, line 303: we still need this part?
we convert to address?

contracts/sol6/KyberMatchingEngine.sol
```diff
@@ -31,14 +31,19 @@ contract KyberMatchingEngine is KyberHintHandler, IKyberMatchingEngine, Withdraw
...
         emit KyberStorageUpdated(_kyberStorage);
         kyberStorage = _kyberStorage;
         return true;
     }
 
-    function setNegligbleRateDiffBps(uint256 _negligibleRateDiffBps) external returns (bool) {
+    function setNegligbleRateDiffBps(uint256 _negligibleRateDiffBps)
+        external
+        virtual
+        override
```
ilanDoron, line 44: what does this override?

contracts/sol6/KyberMatchingEngine.sol
```diff
@@ -69,6 +73,7 @@ contract KyberMatchingEngine is KyberHintHandler, IKyberMatchingEngine, Withdraw
     )
         external
         view
+        override
```
ilanDoron, line 76: what does it override?

buidlerConfigSol4.js
Anyhowclick: Not needed, since we only use this file for compilation.

Anyhowclick: Same for buidlerConfigSol5.

contracts/sol6/IKyberDAO.sol
Anyhowclick: Yeah, didn't know too

contracts/sol6/Dao/KyberFeeHandler.sol
Anyhowclick: I'm not sure when we need fallback() if we have receive()... Not sure what the best practice is here.

contracts/sol6/KyberMatchingEngine.sol
Anyhowclick: Interface

contracts/sol6/KyberHintHandler.sol
Anyhowclick: Hmmmm, since we shift to storage, maybe we can read from there instead?

Anyhowclick: We can discuss this. Cos network side also stores ids to address... if we just get from storage, then we can remove it too.

contracts/sol6/Dao/KyberFeeHandler.sol
Anyhowclick: "Prefer using a receive function only, whenever possible."
https://solidity.readthedocs.io/en/v0.6.6/060-breaking-changes.html#how-to-update-your-code

Will remove

PR #911, opened 2020-04-24
update public facing interfaces in readme file

ilanDoron:
I am trying to be clear for external developers using this repo
which APIs are for public use.

and which are internal

for internal ones we might change every version.
public facing API we should maintain as much as possible.

also fixing some changes related to sol6 changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,12 +2,13 @@
 This repository contains kyber network smart contracts.
 For more details, please visit our [developer portal](https://developer.kyber.network/)
 
-##API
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,6 +2,13 @@
 This repository contains kyber network smart contracts.
 For more details, please visit our [developer portal](https://developer.kyber.network/)
 
+##API
```
Anyhowclick, line 5: Space after ##

PR #923, opened 2020-04-30
Removed rateWithoutFees and getPriceDataNoFees, updated rateWithNetworkFee calculation

Anyhowclick:
Cannot work if maxDestAmount is used though....

### DIFF
--- a/contracts/sol6/IKyberNetwork.sol
+++ b/contracts/sol6/IKyberNetwork.sol
@@ -50,7 +50,6 @@ interface IKyberNetwork {
         external
         view
         returns (
-            uint256 expectedRateNoFees,
             uint256 expectedRateAfterNetworkFee,
             uint256 expectedRateAfterAllFees
         );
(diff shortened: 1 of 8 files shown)


manhlx3006:
Assume user wanted to trade 100 KNC for 1 ETH. Platform fee: 10%, network fee: 20%.

Min acceptable rate: 0.008

SrcAmount = 100 KNC
DestAmountOnlyNetworkFee = 0.8 (1 - 20% * 1)
-> Rate only network fee = 0.008 (>= min acceptable rate -> this trade should pass)

SrcAmountWithPlatformFee = 100 * (1 - 10%) = 90
Delta dest amount = 0.7 (1 - 10% * 1 - 20% * 1)
-> Rate only network fee = 0.7 / 90 = 0.0078 (< min acceptable rate -> this trade will be reverted)

In this case, is it ok for the trade to revert due to minConversionRate?

Note: it is impossible to compute exact result without knowing of network fee percentage.

Anyhowclick:
The trade won't revert, as `rateWithNetworkFee` is now calculated as the 2nd method. Unless the taker platform calculates it like the first method.

It boils down to the expected behaviour we want. From what I understand, the ideal behaviour is that when taker platforms call `getExpectedRateWithHintAndFees` with the platform fee, there shouldn't be a problem using `rateWithNetworkFee` as the minimum conversion rate.

PR #925, opened 2020-05-01
code clean up. comments and unused code

ilanDoron:


contracts/sol5/utils/Utils4.sol
```diff
@@ -6,7 +6,8 @@ import "../IERC20.sol";
 /**
  * @title Kyber utility file
  * mostly shared constants and rate calculation helpers
- * inherited by most of kyber contracts
+ * inherited by most of kyber contracts.
+ * previous utils implementations are for previous solidity version.
```
Anyhowclick, line 10: versions

ilanDoron: but before version 5 we have one version. (for our code)

contracts/sol6/mock/MaliciousKyberNetwork.sol
```diff
@@ -111,23 +111,14 @@ contract MaliciousKyberNetwork is KyberNetwork {
...
-    /// @param destAddress Address to send tokens to
-    /// @return true if trade is successful
     function doReserveTrades(
         IERC20 src,
         uint256 amount,
         IERC20 dest,
         address payable destAddress,
         TradeData memory tData,
         uint256 expectedDestAmount
-    ) internal override returns (bool) {
```
manhlx3006, line 130: Should remove amount here?

contracts/sol6/mock/MaliciousKyberNetwork2.sol
```diff
@@ -1,30 +1,20 @@
...
+        myFeeWei = 10;
     }
 
-    /// @notice use token address ETH_TOKEN_ADDRESS for ether
-    /// @dev do one trade with a reserve
-    /// @param src Src token
-    /// @param amount amount of src tokens
-    /// @param dest   Destination token
-    /// @param destAddress Address to send tokens to
-    /// @return true if trade is successful
```
manhlx3006, line 27: Should remove amount here?

contracts/sol6/mock/MaliciousKyberNetwork.sol
ilanDoron: malicious network is used to send a bit less dest amount. add see the proxy can identify and revert.

so this function is doing it well. don't think we should re write it.

PR #928, opened 2020-05-05
Verify Network balances of src & dst tokens after each reserve trade

manhlx3006:


contracts/sol6/KyberNetwork.sol
```diff
@@ -574,24 +574,24 @@ contract KyberNetwork is WithdrawableNoModifiers, Utils5, IKyberNetwork, Reentra
         ReservesData memory reservesData = (src == ETH_TOKEN_ADDRESS)
             ? tradeData.ethToToken
             : tradeData.tokenToEth;
-        uint256 callValue;
 
-        for (uint256 i = 0; i < reservesData.addresses.length; i++) {
-            callValue = (src == ETH_TOKEN_ADDRESS) ? reservesData.srcAmounts[i] : 0;
+        uint256 srcDecimals = getUpdateDecimals(src);
```
ilanDoron, line 578: do we need update here?
we have setDecimals in listTokenForReserve

manhlx3006: If the decimal is already set for this token, the effect (and properly gas consumption) is the same for `getUpdateDecimals` and `getDecimals` right?

Anyhowclick: I would do it differently, not sure if more gas efficient:
```
   reservesData = tradeData.ethToToken;
   srcDecimals = ETH_DECIMALS;
   destDecimals = reservesData.decimals;
} else {
   reservesData = tradeData.tokenToEth;
   srcDecimals = reservesData.decimals;
   destDecimals = ETH_DECIMALS;
```

ilanDoron: actually looking again.
we could be sending correct reserveData from calling function. don't need tradeData

on same lines
can be sending src / dest Decimals.
don't need to query from storage.
And amount can be removed if stack too deep (its being removed in another PR)

manhlx3006: Updated

contracts/sol6/KyberNetwork.sol
```diff
@@ -1184,16 +1241,20 @@ contract KyberNetwork is WithdrawableNoModifiers, Utils5, IKyberNetwork, Reentra
...
+                // revert back to use current src amounts
+                newSrcAmount = srcAmount;
+                return newSrcAmount;
+            }
 
-            reservesData.srcAmounts[i] = newSrcAmount;
-            srcAmount += newSrcAmount;
+            newSrcAmount += newSrcAmounts[i];
         }
+        reservesData.srcAmounts = newSrcAmounts;
```
ilanDoron, line 1258: maybe add comment. 
// new src amounts used only any of them isn't greater then current srcAmount.

contracts/sol6/KyberNetwork.sol
```diff
@@ -601,6 +597,62 @@ contract KyberNetwork is WithdrawableNoModifiers, Utils5, IKyberNetwork, Reentra
...
+                address(this),
+                conversionRate,
+                true
+            ),
+            "trade failed"
+        );
+
+        uint256 balanceAfter = getBalance(src, address(this));
+        // verify correct src amount is taken
+        require(balanceAfter <= srcBalanceBefore);
```
ilanDoron, line 645: can have both requires as one and 1 message?

contracts/sol6/KyberNetwork.sol
```diff
@@ -571,27 +571,23 @@ contract KyberNetwork is WithdrawableNoModifiers, Utils5, IKyberNetwork, Reentra
...
+                src,
+                reservesData.srcAmounts[i],
+                dest,
+                reservesData.rates[i],
+                srcDecimals,
+                destDecimals
             );
         }
+        // verify balance after all reserve trades
+        require(actualDestAmount >= expectedDestAmount, "amt low");
```
ilanDoron, line 590: Can we avoid the actualDestAmount calculation and compare?
since if each reserve sends correct amount. it seems like extra check

contracts/sol6/KyberNetwork.sol
```diff
@@ -1105,13 +1166,14 @@ contract KyberNetwork is WithdrawableNoModifiers, Utils5, IKyberNetwork, Reentra
         }
 
         // reverse calculation, because we are working backwards
-        tradeData.tradeWei =
+        uint256 newTradeWei =
```
ilanDoron, line 1169: can remove from this PR?

contracts/sol6/KyberNetwork.sol
manhlx3006: Yes, it is removed, will push the changes later along with some tests

contracts/sol6/KyberNetwork.sol
manhlx3006: Yeah I rebased from Katalyst

contracts/sol6/KyberNetwork.sol
```diff
@@ -594,6 +590,66 @@ contract KyberNetwork is WithdrawableNoModifiers, Utils5, IKyberNetwork, Reentra
...
+                src,
+                srcAmount,
+                dest,
+                address(this),
+                conversionRate,
+                true
+            ),
+            "trade failed"
+        );
+
```
ilanDoron, line 635: please use amount and not amt

its not a common abbreviation.

Also I notice that src calc is reuqired only if src isn't ETH.
For Eth we know how much source was sent.

contracts/sol6/KyberNetwork.sol
```diff
@@ -594,6 +590,66 @@ contract KyberNetwork is WithdrawableNoModifiers, Utils5, IKyberNetwork, Reentra
...
+        uint256 balanceAfter = getBalance(src, address(this));
+        // verify correct src amount is taken
+        if (srcBalanceBefore >= balanceAfter && srcBalanceBefore - balanceAfter > srcAmount) {
+            revert("src amt high");
+        }
+        // verify correct dest amount is received
+        balanceAfter = getBalance(dest, address(this));
+        if (balanceAfter < destBalanceBefore || balanceAfter - destBalanceBefore < expectedDestAmount) {
```
ilanDoron, line 643: please use amount

contracts/sol6/KyberNetwork.sol
manhlx3006: yeah, for eth we won't check for src balance of network before and after trade

PR #935, opened 2020-05-06
add overflow check at calcTradeSrcAmount

ducquangkstn:
This PR should wait till we finalize overflow session
any overflow checks or comments should be added here.


contracts/sol6/KyberNetwork.sol
```diff
@@ -809,6 +809,7 @@ contract KyberNetwork is WithdrawableNoModifiers, Utils5, IKyberNetwork, Reentra
             hint
         );
 
+        require(tradeData.tradeWei <= MAX_QTY, "Trade wei > MAX_QTY");
```
Anyhowclick, line 812: Recommend small caps: Trade -> trade

ilanDoron: Oh
Missed this.

PR #936, opened 2020-05-08
Fix reverse amount calculation for trades. When maxDestAmount is small and we recalculate trade values in reverse.

manhlx3006:
When doing reverse calculation:
- new src amounts are used only when any of them isn't greater then corresponding current srcAmount.
- If newTradeWei > tradeWei, fallback to use tradeWei.

### DIFF
--- a/contracts/sol6/KyberNetwork.sol
+++ b/contracts/sol6/KyberNetwork.sol
@@ -1095,6 +1095,7 @@ contract KyberNetwork is WithdrawableNoModifiers, Utils5, IKyberNetwork, Reentra
         uint256 weiAfterDeductingFees;
         if (tradeData.input.dest != ETH_TOKEN_ADDRESS) {
             weiAfterDeductingFees = calcTradeSrcAmount(
+                tradeData.tradeWei - tradeData.platformFeeWei - tradeData.networkFeeWei,
                 ETH_DECIMALS,
                 tradeData.ethToToken.decimals,
                 tradeData.input.maxDestAmount,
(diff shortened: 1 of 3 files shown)


contracts/sol6/KyberNetwork.sol
```diff
@@ -1133,13 +1134,14 @@ contract KyberNetwork is WithdrawableNoModifiers, Utils5, IKyberNetwork, Reentra
         }
 
         // reverse calculation, because we are working backwards
-        tradeData.tradeWei =
+        uint256 newTradeWei =
             (weiAfterFees * BPS * BPS) /
```
ilanDoron, line 1138: maybe we should still improve calculation here.
like mul with BPS ** 4

anyway its better to mul with constant value. and avoid const * const
seems the compiler doesn't do it in compilation time...

Anyhowclick: My concern is that multiplying with that large a number might cause overflow problems

Anyhowclick: I guess it depends what's the maximum theoretical value of `calcTradeSrcAmount`?

ilanDoron: OK
seems to me we should not update this value.
worst case is a few wei error.
we can keep existing numbers.

contracts/sol6/KyberNetwork.sol
```diff
@@ -1140,37 +1143,47 @@ contract KyberNetwork is WithdrawableNoModifiers, Utils5, IKyberNetwork, Reentra
...
                 destAmountSplit,
                 srcDecimals,
                 destDecimals,
                 reservesData.rates[i]
             );
-            newSrcAmount = minOf(newSrcAmount, currentSrcAmount);
+            if (newSrcAmounts[i] > currentSrcAmount) {
+                // revert back to use current src amounts
+                newSrcAmount = srcAmount;
+                return newSrcAmount;
```
Anyhowclick, line 1181: Will it work if u just `return srcAmount` instead? o.O

PR #940, opened 2020-05-11
Add abstraction layer for network, feeHandler, DAO and matchingEngine history

Anyhowclick:


ducquangkstn:
I think we could create history as mapping (bytes32 => []address) like
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/AccessControl.sol
So we don't have to deploy `networkHistory` `feeHandlerHistory` `kyberDAOHistory` and `matchingEngineHistory`

Anyhowclick:
We can, but we want this additional abstraction layer as it will help with upgradeability.

In the case where we want to upgrade the storage contract to store more contract types (like the staking contracts for example), then we don't have to re-initialise and do migration.

contracts/sol6/KyberStorage.sol
```diff
@@ -66,10 +67,28 @@ contract KyberStorage is IKyberStorage, PermissionGroupsNoModifiers, Utils5 {
         onlyAdmin();
         require(_kyberNetwork != IKyberNetwork(0), "network 0");
         emit KyberNetworkUpdated(_kyberNetwork);
-        previousNetworks.push(kyberNetwork);
+        networkHistory.saveContract(address(_kyberNetwork));
         kyberNetwork = _kyberNetwork;
     }
 
+    function setHistoricalContracts(
```
ilanDoron, line 74: seems to me this better be called init

also its a one time init.
right?
So better comment that.

contracts/sol6/KyberStorage.sol
```diff
@@ -66,10 +67,28 @@ contract KyberStorage is IKyberStorage, PermissionGroupsNoModifiers, Utils5 {
...
+        IKyberHistory _kyberDAOHistory,
+        IKyberHistory _matchingEngineHistory
+    ) external {
+        onlyAdmin();
+        require(_networkHistory != IKyberHistory(0), "networkHistory 0");
+        require(_feeHandlerHistory != IKyberHistory(0), "feeHandlerHistory 0");
+        require(_kyberDAOHistory != IKyberHistory(0), "kyberDAOHistory 0");
+        require(_matchingEngineHistory != IKyberHistory(0), "matchingEngineHistory 0");
+
+        networkHistory = _networkHistory;
```
ilanDoron, line 86: seems like this should be a one time call
I mean.
what motivation do we have to replace if it already stores some values.

if we do replace it.
we loose some trust for this list - from reserves / users.

contracts/sol6/KyberStorage.sol
```diff
@@ -81,39 +100,22 @@ contract KyberStorage is IKyberStorage, PermissionGroupsNoModifiers, Utils5 {
...
-            feeHandler.push(_feeHandler);
-        }
-
-        if (matchingEngine.length > 0) {
-            matchingEngine.push(matchingEngine[0]);
-            matchingEngine[0] = newMatchingEngine;
-        } else {
-            matchingEngine.push(newMatchingEngine);
-        }
+        feeHandlerHistory.saveContract(_feeHandler);
```
ilanDoron, line 111: if history contract isn't set will revert in a non elegant way.
this is what u meant?

Its not that bad.
but we might want to issue some uninitialized revert message.
maybe both can have the same message.

contracts/sol6/KyberStorage.sol
```diff
@@ -81,39 +100,22 @@ contract KyberStorage is IKyberStorage, PermissionGroupsNoModifiers, Utils5 {
...
         onlyNetwork();
-        
-        if (kyberDAO.length > 0) {
-            kyberDAO.push(kyberDAO[0]);
-            kyberDAO[0] = _kyberDAO;
-        } else {
-            kyberDAO.push(_kyberDAO);
-        }
+
+        kyberDAOHistory.saveContract(_kyberDAO);
```
ilanDoron, line 118: same here for uninitialized history contract DB

test/sol6/kyberStorage.js
```diff
@@ -680,7 +794,22 @@ contract('KyberStorage', function(accounts) {
...
+            networkHistory = await KyberHistory.new(admin);
+            feeHandlerHistory = await KyberHistory.new(admin);
+            kyberDAOHistory = await KyberHistory.new(admin);
+            matchingEngineHistory = await KyberHistory.new(admin);
             let mockStorage = await MockStorage.new(admin);
+            await networkHistory.setStorageContract(mockStorage.address, {from: admin});
+            await feeHandlerHistory.setStorageContract(mockStorage.address, {from: admin});
+            await kyberDAOHistory.setStorageContract(mockStorage.address, {from: admin});
+            await matchingEngineHistory.setStorageContract(mockStorage.address, {from: admin});
+            await mockStorage.setHistoricalContracts(
```
ilanDoron, line 806: don't see a test where this isn't initialized.

contracts/sol6/KyberStorage.sol
Anyhowclick: Yes right. In that case, will shift to constructor.

Anyhowclick:
Shifting to ctor arguments should solve all the issues u mentioned

PR #941, opened 2020-05-11
sol6 new features. set user defined types in mapping. Add immutable type.

Anyhowclick:


contracts/sol6/Dao/KyberDAO.sol
```diff
@@ -57,9 +57,9 @@ contract KyberDAO is IKyberDAO, EpochUtils, ReentrancyGuard, Utils5, DaoOperator
 
     // minimum duration in seconds for a campaign
     uint256 public minCampaignDurationInSeconds = 345600; // around 4 days
-    IERC20 public kncToken;
-    KyberStaking public staking;
-    IKyberFeeHandler public feeHandler;
+    IERC20 public immutable kncToken;
+    IKyberStaking public staking;
```
ducquangkstn, line 61: After PR deploying staking at Dao constructor, staking could be marked immutable

contracts/sol6/KyberStorage.sol
```diff
@@ -191,12 +191,12 @@ contract KyberStorage is IKyberStorage, PermissionGroupsNoModifiers, Utils5 {
 
         // update reserve mappings
         reserveIdToAddresses[reserveId].push(reserveIdToAddresses[reserveId][0]);
-        reserveIdToAddresses[reserveId][0] = address(0);
-        reserveAddressToId[reserve] = bytes32(0);
+        delete reserveIdToAddresses[reserveId][0];
```
ducquangkstn, line 194: It seems that the logic in here is not tested yet

ilanDoron: actually this part is wrong
will reverse this change.

contracts/sol6/Dao/KyberDAO.sol
ilanDoron: right
will update

Anyhowclick: One other reason why it wasn't set to immutable is because the mock contract allows it to change, but on closer inspection, this can (and has) been fixed.

Same for the dao contract. See latest commit.

PR #943, opened 2020-05-11
remove only return true function

ductm54:


test/sol6/kyberNetwork.js
```diff
@@ -2074,53 +2073,6 @@ contract('KyberNetwork', function(accounts) {
         });
     });
 
-    describe("test handle change edge case", async function(){
```
ilanDoron, line 2077: why these tests removed?

ductm54: no need to mock to test when required is delete

ilanDoron:
But these tests are still important.

Seems they test we handle this edge case correctly correctly
Don't follow why it should be removed.

contracts/sol6/Dao/KyberDAO.sol
```diff
@@ -437,10 +437,7 @@ contract KyberDAO is IKyberDAO, EpochUtils, ReentrancyGuard, Utils5 {
 
         hasClaimedReward[staker][epoch] = true;
         // call fee handler to claim reward
-        require(
-            feeHandler.claimStakerReward(staker, perInPrecision, epoch),
-            "claimReward: feeHandle failed to claim"
-        );
+        feeHandler.claimStakerReward(staker, perInPrecision, epoch);
```
manhlx3006, line 440: May be we can change FeeHandler to:
`require(success, "staker rewards transfer failed");`
---> ` if (!success) { return false; }` and emit event

ilanDoron: for readability?

test/sol6/kyberDao.js
```diff
@@ -3586,35 +3585,6 @@ contract('KyberDAO', function(accounts) {
       await feeHandler.withdrawAllETH({from: accounts[0]});
     });
 
-    it("Test claim reward should revert fee handler return false for claimReward", async function() {
```
manhlx3006, line 3589: Should still keep this test, but change to: "Test claim reward should revert if fee handler failed to claim reward` and make it revert in FeeHandler `claimStakerReward` function

contracts/sol6/Dao/KyberDAO.sol
manhlx3006: Yeah mostly for readability. Not sure if we should catch this event and notify staker that this address can not receive ETH

ilanDoron: so idea is to add an event in feeHander.

Anyhowclick: IMO leave it like that.. feeHandler should revert if the failing msg

PR #945, opened 2020-05-12
Verify invalid, increasing order, or unregistered reserveIds for all external hint functions

ayobuenavista:


contracts/sol6/KyberHintHandler.sol
```diff
@@ -253,11 +296,50 @@ abstract contract KyberHintHandler is IKyberHint, Utils5 {
...
-    /// @notice Ensures that the reserveIds passed when building hints are in increasing sequence
+    /// @notice Checks if the reserveId exists
+    /// @param reserveId Reserve IDs array
+    function checkReserveIdsExists(bytes32 reserveId)
+        internal
+        view
+    {
+        address[] memory reserveAddresses;
+
+        reserveAddresses = getReserveAddresses(reserveId);
```
ilanDoron, line 307: why not use getReserveAddress we already have?
it doesn't work for this check?

contracts/sol6/KyberHintHandler.sol
```diff
@@ -253,11 +296,49 @@ abstract contract KyberHintHandler is IKyberHint, Utils5 {
         (t2eHint, e2tHint) = abi.decode(hint, (bytes, bytes));
     }
 
-    /// @notice Ensures that the reserveIds passed when building hints are in increasing sequence
+    /// @notice Checks if the reserveId exists
+    /// @param reserveId Reserve IDs array
```
Anyhowclick, line 300: Wrong param comment...

ayobuenavista: Fixed

PR #946, opened 2020-05-12
[WIP] Add dgx tests

ducquangkstn:


test/sol6/kyberNetwork.js
```diff
@@ -2180,6 +2182,137 @@ contract('KyberNetwork', function(accounts) {
...
+            let dgxDecimal = new BN(9);
+            let networkProxy = accounts[8];
+            let tokens;
+            before("setup, add and list mock reserves", async() => {
+                let dgxStorage = await DummyDGXStorage.new({from: admin});
+                dgxToken = await DummyDGX.new(dgxStorage.address, admin);
+                await dgxStorage.setInteractive(dgxToken.address, { from: admin});
+                // transfer token and add accounts[0] to whitelist so `token.transfer` still works
+                await dgxToken.mintDgxFor(accounts[0], new BN(10).pow(new BN(18)), {from: admin});
+                await dgxToken.updateUserFeesConfigs(accounts[0], true, true, {from: admin});
```
ilanDoron, line 2198: what is accounts[0]
easier to read if using the name - admin? trader?

test/sol6/kyberNetwork.js
```diff
@@ -2180,6 +2182,137 @@ contract('KyberNetwork', function(accounts) {
...
+                let hintType = EMPTY_HINTTYPE;
+                hint = await nwHelper.getHint(rateHelper, matchingEngine, reserveInstances, hintType, undefined, dgxToken.address, ethAddress, srcQty);
+
+                info = [srcQty, networkFeeBps, platformFeeBps];
+
+                expectedResult = await nwHelper.getAndCalcRates(matchingEngine, storage, reserveInstances,
+                    dgxToken.address, ethAddress, srcQty,
+                    dgxDecimal, ethDecimals,
+                    networkFeeBps, platformFeeBps, hint);
+                await dgxToken.mintDgxFor(networkProxy, srcQty, {from: admin});
```
ilanDoron, line 2265: Oh
I understand proxy is the source address of the trade here. so its "taker"

seems to me better if adding a real proxy here 
(even though its network file.) we can add it since essentially we are testing the network.
And with proxy we test real flow.

test/sol6/kyberNetwork.js
ducquangkstn: For TestToken, we call the constructor at accounts[0] and then we transfer token to reserve at networkHelper
https://github.com/KyberNetwork/smart-contracts/blob/6b45af0be60e0d6cc4566d222965e566930f68fc/test/sol6/networkHelper.js#L103
In this transaction, from address is not specified so truffle will use default address (accounts[0)
So to re-use the setupReserve code. We need to add supply and whitelist accounts[0]

test/sol6/kyberNetwork.js
ducquangkstn: change as request

test/sol6/kyberNetwork.js
ilanDoron: OK
got it.

PR #950, opened 2020-05-12
[don't pull] overflow comments

ilanDoron:
overflow comments for reviewing our logic.

//+ is where we can see value limit for some number.
//* is when we have to review this math. and approve it is ok


contracts/sol6/KyberNetwork.sol
```diff
@@ -865,10 +867,22 @@ contract KyberNetwork is WithdrawableNoModifiers, Utils5, IKyberNetwork, Reentra
...
     ) internal view returns (uint256[] memory feesAccountedDestBps) {
         uint256 numReserves = reservesData.ids.length;
+        uint256 srcAmountAfterFee;
+        uint256 destAmountFeeBps;
+
+        if (src == ETH_TOKEN_ADDRESS) {
+            //+ srcAmountAfterFee < MAX_QTY
+            srcAmountAfterFee = srcAmount - 
+                (tradeData.tradeWei * tradeData.networkFeeBps / BPS);
+                require(srcAmountAfterFee < srcAmount, "high fees");
```
ducquangkstn, line 880: should we change the msg to `underflow subtraction`. And add a comment that it is a safe check ( can not reach from our code)

contracts/sol6/KyberNetwork.sol
```diff
@@ -881,16 +895,11 @@ contract KyberNetwork is WithdrawableNoModifiers, Utils5, IKyberNetwork, Reentra
...
-                if (src == ETH_TOKEN_ADDRESS) {
-                    // reduce fee from srcAmount if fee paying.
-                    reservesData.srcAmounts[i] =
-                        ((srcAmount - networkFeeValue) * reservesData.splitsBps[i]) /
-                        BPS;
-                } else {
-                    reservesData.srcAmounts[i] = (srcAmount * reservesData.splitsBps[i]) / BPS;
-                    feesAccountedDestBps[i] = networkFeeBps;
-                }
+                reservesData.srcAmounts[i] = srcAmountAfterFee * reservesData.splitsBps[i] / BPS;
```
ducquangkstn, line 901: Can you explain the reason why move the  calculation outside of the if block

contracts/sol6/KyberNetwork.sol
```diff
@@ -758,11 +758,13 @@ contract KyberNetwork is WithdrawableNoModifiers, Utils5, IKyberNetwork, Reentra
             hint
         );
 
+        require(tradeData.tradeWei <= MAX_QTY, "tradeWei > MAX_QTY");
         if (tradeData.tradeWei == 0) {
             return (0, 0);
         }
 
         // platform fee
+        //*
```
ducquangkstn, line 767: max(tradeData.tradeWei) = max(validateTradeCalcDestQtyAndFeeData) = MAX_RATE *MAX_QTY * 10 ** (MAX_DECIMAL - ETH_DECIMAL) / PRECISION * BPS =1e57

contracts/sol6/KyberNetwork.sol
ilanDoron: srcAmountAfterFee is calculated for src = ETH_TOKEN_ADDRESS
so when ETH_TOKEN_ADDRESS we always subtract networkFeeValue. so just did one subtraction.
if src != ETH_TOKEN_ADDRESS it will just be srcAmount.
So should keep same behavior

contracts/sol6/KyberNetwork.sol
ilanDoron: removing message

contracts/sol6/KyberNetwork.sol
```diff
@@ -758,11 +758,13 @@ contract KyberNetwork is WithdrawableNoModifiers, Utils5, IKyberNetwork, Reentra
             hint
         );
 
+        require(tradeData.tradeWei <= MAX_QTY, "tradeWei > MAX_QTY");
```
Anyhowclick, line 761: Ethereum's total supply is 111 million = `1e26`...

IMO this check is somewhat unnecessary, because reserves won't hold that much ETH to transfer to the network, so the trade will revert anyway.

ilanDoron: Yes

but it helps later when calculating reverse trade.
can have clear assumption on quantity

contracts/sol6/KyberNetwork.sol
ilanDoron: can resolve conversation?

contracts/sol6/KyberNetwork.sol
```diff
@@ -809,12 +813,14 @@ contract KyberNetwork is WithdrawableNoModifiers, Utils5, IKyberNetwork, Reentra
             hint
         );
 
+        //* circualting eth 1e26 = 110 Million ETH. but we need this limit assumption for reverse calc.
```
Anyhowclick, line 816: circulating

contracts/sol6/KyberNetwork.sol
```diff
@@ -956,6 +974,7 @@ contract KyberNetwork is WithdrawableNoModifiers, Utils5, IKyberNetwork, Reentra
                 reservesData.srcAmounts[i],
                 block.number
             );
+            //* reservesData.rates[i] <= MAX_RATE?  later validation when calc destAmount
```
Anyhowclick, line 977: //* reservesData.rates[i] <= MAX_RATE. validated in validateTradeCalcDestQtyAndFeeData when calcDstQty is called

contracts/sol6/KyberNetwork.sol
```diff
@@ -1027,6 +1047,7 @@ contract KyberNetwork is WithdrawableNoModifiers, Utils5, IKyberNetwork, Reentra
         require(input.srcAmount != 0, "0 srcAmt");
         require(input.destAddress != address(0), "dest add 0");
         require(input.src != input.dest, "src = dest");
+        //+ platformFee and NetworkFee limits
```
Anyhowclick, line 1050: networkFee

contracts/sol6/KyberNetwork.sol
```diff
@@ -809,12 +813,14 @@ contract KyberNetwork is WithdrawableNoModifiers, Utils5, IKyberNetwork, Reentra
...
         );
 
+        //* circualting eth 1e26 = 110 Million ETH. but we need this limit assumption for reverse calc.
         require(tradeData.tradeWei <= MAX_QTY, "Trade wei > MAX_QTY");
         if (tradeData.tradeWei == 0) {
             return (0, 0);
         }
 
         // calculate fees
+        //* networkFeeBps * 2 + platfromFeeBps < BPS
```
manhlx3006, line 823: may be added: `feeAccountedBps <= 2 * BPS`?

contracts/sol6/KyberNetwork.sol
```diff
@@ -1105,24 +1126,28 @@ contract KyberNetwork is WithdrawableNoModifiers, Utils5, IKyberNetwork, Reentra
...
 
+            //+ destAmount <= MAX_QTY * MAX_RATE == 1e53
             destAmount += calcDstQty(
                 reservesData.srcAmounts[i],
                 srcDecimals,
                 destDecimals,
                 reservesData.rates[i]
             );
 
+            //+ feeAccountedBps <= BPS
```
manhlx3006, line 1145: Should add comment that this is only one trade, so `feeAccountedBps <= BPS`, since other places we comment that `feeAccountedBps <= 2 * BPS`

ilanDoron: Oh yes. right.

contracts/sol6/KyberNetwork.sol
ilanDoron: OK
I only now understand this comment.
it means we have 10k reserves. each has 1 bps split. right?

ilanDoron:
closing in favour of a new PR with updated base code.

PR #955, opened 2020-05-15
Move loop trade and verify balances of reserves into one function

manhlx3006:


contracts/sol6/KyberNetwork.sol
```diff
@@ -579,71 +575,61 @@ contract KyberNetwork is WithdrawableNoModifiers, Utils5, IKyberNetwork, Reentra
...
         IERC20 src,
-        uint256 srcAmount,
         IERC20 dest,
-        uint256 conversionRate,
         uint256 srcDecimals,
         uint256 destDecimals
     ) internal
     {
+        // only need to verify src balance if src is not ETH
+        uint256 srcBalanceBefore = (src == ETH_TOKEN_ADDRESS) ? 0 : getBalance(src, address(this));
```
Anyhowclick, line 593: can remove (src == ETH_TOKEN_ADDRESS)
`uint256 srcBalanceBefore = getBalance(src, address(this));`

manhlx3006: if src token is ETH, we don't need to verify src token balance as we will pass this value to reserve's trade function

Anyhowclick: Yes, so it should be the same as the line below right?
`uint256 srcBalanceBefore = (src == ETH_TOKEN_ADDRESS) ? 0 : getBalance(src, address(this));`
should be
`uint256 srcBalanceBefore = getBalance(src, address(this));`
like the line below?

manhlx3006: if we don't need to verify src token balance, why do we need to get network's balance of src token?

Anyhowclick: Ahhhhhhhh righhtttt, ok got it

manhlx3006:
Bytecode size only increases 4 compare to current Katalyst branch

contracts/sol6/KyberNetwork.sol
```diff
@@ -579,71 +581,61 @@ contract KyberNetwork is WithdrawableNoModifiers, Utils5, IKyberNetwork, Reentra
...
+            require(
+                reservesData.addresses[i].trade{value: callValue}(
+                    src,
+                    reservesData.srcAmounts[i],
+                    dest,
+                    address(this),
+                    reservesData.rates[i],
+                    true
+                ),
+                "trade failed"
```
ilanDoron, line 613: maybe better
reserve trade failed

so its different from 
kyber trade failed.

test/sol6/kyberNetwork.js
```diff
@@ -612,38 +612,30 @@ contract('KyberNetwork', function(accounts) {
...
-                    "reserve list is empty"
-                )
+                let tx = await tempNetwork.listReservesForToken(
+                    newToken.address, 0, 0, true, {from: operator}
+                );
+                // no event should be emitted
+                Helper.assertEqual(tx.receipt.logs.length, 0);
             });
 
             it("test list reserves empty as wrong indices", async() => {
```
ilanDoron, line 631: update title.

test/sol6/kyberNetwork.js
```diff
@@ -612,38 +612,30 @@ contract('KyberNetwork', function(accounts) {
...
+                    newToken.address, 4, 5, true, {from: operator}
                 )
-                await expectRevert(
-                    tempNetwork.listReservesForToken(
-                        newToken.address, 4, 4, true, {from: operator}
-                    ),
-                    "reserve list is empty"
+                await tempNetwork.listReservesForToken(
+                    newToken.address, 4, 4, true, {from: operator}
                 )
```
ilanDoron, line 639: also here should check no event.
no?

contracts/sol6/KyberNetwork.sol
manhlx3006: updated

test/sol6/kyberNetwork.js
manhlx3006: updated

test/sol6/kyberNetwork.js
manhlx3006: updated

PR #959, opened 2020-05-18
Ensure knc token is the same in Staking and DAO contracts

manhlx3006:


contracts/sol6/Dao/KyberStaking.sol
```diff
@@ -18,7 +17,7 @@ contract KyberStaking is IKyberStaking, EpochUtils, ReentrancyGuard {
         address delegatedAddress;
     }
 
-    IERC20 public kncToken;
+    IERC20 public override kncToken;
```
ilanDoron, line 20: why not knc?

I think it should be clear enough.

test/sol6/kyberDao.js
```diff
@@ -5850,6 +5850,17 @@ contract('KyberDAO', function(accounts) {
...
+      // knc is different from staking
+      let token = await TestToken.new("test token", "tst", 18);
+      await expectRevert(
+        DAOContract.new(
+          blocksToSeconds(10), blockToTimestamp(currentBlock + 10),
+          stakingContract.address,  feeHandler.address, token.address,
+          minCampPeriod, defaultNetworkFee, defaultRewardBps, defaultRebateBps,
+          campCreator
+        ),
+        "ctor: different knc token"
```
ilanDoron, line 5862: maybe
"different knc address"
?

manhlx3006:
Moving to deploy Staking from Dao contract

PR #960, opened 2020-05-19
Enable updating network proxy and setting network address in FeeHandler

manhlx3006:
FeeHandler:
- Enable updating network proxy.
- Setting network address without proxy.

### DIFF
--- a/contracts/sol6/Dao/KyberFeeHandler.sol
+++ b/contracts/sol6/Dao/KyberFeeHandler.sol
@@ -34,9 +34,7 @@ import "./DaoOperator.sol";
  *          Expiry timestamp for data is set. when data expires. Fee handler reads new data from DAO.
  */
 
-interface IKyberProxy is IKyberNetworkProxy, ISimpleKyberProxy {
-    function kyberNetwork() external view returns (address);
-}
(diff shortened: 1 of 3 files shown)


contracts/sol6/Dao/KyberFeeHandler.sol
```diff
@@ -34,9 +34,7 @@ import "./DaoOperator.sol";
  *          Expiry timestamp for data is set. when data expires. Fee handler reads new data from DAO.
  */
 
-interface IKyberProxy is IKyberNetworkProxy, ISimpleKyberProxy {
-    function kyberNetwork() external view returns (address);
-}
+interface IKyberProxy is ISimpleKyberProxy, IKyberNetworkProxy { }
```
Anyhowclick, line 37: Do we have a standard for empty code blocks? With or without spacing?

manhlx3006: saw few places it is `contract A ... {}`

ilanDoron: let me merge
and try to align those later.

PR #961, opened 2020-05-19
Using KyberDao to deploy KyberStaking contract

manhlx3006:


contracts/sol6/Dao/KyberDAO.sol
```diff
@@ -139,6 +132,15 @@ contract KyberDAO is IKyberDAO, EpochUtils, ReentrancyGuard, Utils5, DaoOperator
...
         });
+
+        // deploy staking contract 
+        staking = new KyberStaking({
+            _kncToken: _knc,
+            _epochPeriod: _epochPeriod,
+            _startTimestamp: _startTimestamp,
+            _daoContract: address(this)
+        });
+        emit KyberStakingDeployed(address(staking));
```
ilanDoron, line 143: Don't think an event is required here.
anyway the address can be easily fetched. 
And will never have another one

contracts/sol6/Dao/mock/MockKyberDaoMoreGetters.sol
```diff
@@ -33,7 +31,7 @@ contract MockKyberDaoMoreGetters is KyberDAO {
     }
 
     function replaceStakingContract(address _staking) public {
```
ilanDoron, line 33: function name not clear.
maybe fetchStakingCotractAddress

manhlx3006: oh it is for changing staking contract to be a normal address, so we can call some functions in DAO that require sender is staking.

contracts/sol6/Dao/KyberDAO.sol
manhlx3006: ok, can remove event, staking is public getter anyway

contracts/sol6/Dao/KyberStaking.sol
```diff
@@ -39,47 +37,17 @@ contract KyberStaking is IKyberStaking, EpochUtils, ReentrancyGuard {
         address _kncToken,
         uint256 _epochPeriod,
         uint256 _startTimestamp,
-        address _daoContractSetter
```
ilanDoron, line 42: we can directly take this parameters as IKyberDAO
and above for knc
IERC20
And avoid the casting later.

 I think this is what we usually do.

contracts/sol6/Dao/mock/MockKyberDaoMoreGetters.sol
ilanDoron: Oh
OK

PR #964, opened 2020-05-21
Make sure dest qty of each reserve trade is positive

manhlx3006:


contracts/sol6/KyberNetwork.sol
```diff
@@ -1112,12 +1112,14 @@ contract KyberNetwork is WithdrawableNoModifiers, Utils5, IKyberNetwork, Reentra
             }
             totalBps += reservesData.splitsBps[i];
 
-            destAmount += calcDstQty(
+            uint256 destQty = calcDstQty(
```
ilanDoron, line 1115: lets use destAmount and totalDestAmount
so we are consistent.

in general lets use amount and not qty in all new things we add.

maybe later replace old ones as well.

test/sol6/kyberNetwork.js
```diff
@@ -2947,6 +2949,40 @@ contract('KyberNetwork', function(accounts) {
                 "rate < min rate"
             );
         });
+
+        it("trade dest qty == 0", async function() {
```
ilanDoron, line 2953: please use title

check that when one reserve has zero qty. getExpected rate returns 0. trade reverts

manhlx3006: Updated

contracts/sol6/KyberNetwork.sol
manhlx3006: Updated

contracts/sol6/KyberNetwork.sol
```diff
@@ -1112,12 +1112,16 @@ contract KyberNetwork is WithdrawableNoModifiers, Utils5, IKyberNetwork, Reentra
...
+            uint256 destAmount = calcDstQty(
                 reservesData.srcAmounts[i],
                 srcDecimals,
                 destDecimals,
                 reservesData.rates[i]
             );
+            if (destAmount == 0) {
+                return 0;
+            }
+            totalDestAmount +=  destAmount;
```
Anyhowclick, line 1124: double spacing here

manhlx3006: Fixed

PR #966, opened 2020-05-25
Auto delist tokens when removing reserve

Anyhowclick:
- Mappings `srcTokensPerReserve` and `destTokensPerReserve`
- `getListedTokensByReserveId()` and `getListedTokensByReserveAddress()`
- `delistTokensOfReserve` to automatically delist all tokens when removing reserve

### DIFF
--- a/contracts/sol6/KyberStorage.sol
+++ b/contracts/sol6/KyberStorage.sol
@@ -30,6 +30,8 @@ contract KyberStorage is IKyberStorage, PermissionGroupsNoModifiers, Utils5 {
     mapping(address => bytes32) internal reserveAddressToId;
     mapping(address => bytes32[]) internal reservesPerTokenSrc; // reserves supporting token to eth
     mapping(address => bytes32[]) internal reservesPerTokenDest; // reserves support eth to token
+    mapping(bytes32 => IERC20[]) internal srcTokensPerReserve;
+    mapping(bytes32 => IERC20[]) internal destTokensPerReserve;
 
     uint256 internal feeAccountedPerType = 0xffffffff;
     uint256 internal entitledRebatePerType = 0xffffffff;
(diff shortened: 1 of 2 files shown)


contracts/sol6/KyberStorage.sol
```diff
@@ -585,6 +633,24 @@ contract KyberStorage is IKyberStorage, PermissionGroupsNoModifiers, Utils5 {
             // if reserve wasn't found add it
             reserveArr.push(reserveId);
         }
+
+        for (i = 0; i < tokensArr.length; i++) {
+            if (token == tokensArr[i]) {
+                if (add) {
+                    break; // already added
```
ilanDoron, line 640: regarding this can previous same code.

if already added. can we exit this function?

so this question actually holds for line 622

if we can current loop is enough for remove only option.

Anyhowclick: Mmm yes, it makes sense

contracts/sol6/KyberStorage.sol
```diff
@@ -585,6 +633,22 @@ contract KyberStorage is IKyberStorage, PermissionGroupsNoModifiers, Utils5 {
             // if reserve wasn't found add it
             reserveArr.push(reserveId);
         }
+
+        for (i = 0; i < tokensArr.length; i++) {
```
ilanDoron, line 637: I think here !add should be the external one.
so we don't loop if not removing.

ilanDoron: so actually if add
just push (don't check tokensArr length.

Else
do the loop and remove

contracts/sol6/KyberStorage.sol
```diff
@@ -501,6 +506,32 @@ contract KyberStorage is IKyberStorage, PermissionGroupsNoModifiers, Utils5 {
...
+        returns (
+            IERC20[] memory srcTokens,
+            IERC20[] memory destTokens
+        )
+    {
+        srcTokens = srcTokensPerReserve[reserveId];
+        destTokens = destTokensPerReserve[reserveId];
+    }
+
+    function getListedTokensByReserveAddress(address reserve)
```
ilanDoron, line 521: can remove this.
if not used in trade

PR #967, opened 2020-05-25
remove require always true

ductm54:


contracts/sol6/Dao/KyberDAO.sol
```diff
@@ -745,8 +742,6 @@ contract KyberDAO is IKyberDAO, EpochUtils, ReentrancyGuard, Utils5, DaoOperator
         require(cInPrecision <= POWER_128, "validateParams: c is high");
 
         require(tInPrecision <= POWER_128, "validateParams: t is high");
-
-        return true;
```
manhlx3006, line 749: this is a public function for our operation team to check if params are valid. Not sure if we should remove return here?

contracts/sol6/Dao/KyberDAO.sol
```diff
@@ -775,7 +770,7 @@ contract KyberDAO is IKyberDAO, EpochUtils, ReentrancyGuard, Utils5, DaoOperator
     /**
      * @dev options are indexed from 1
      */
-    function validateVoteOption(uint256 campaignID, uint256 option) internal view returns (bool) {
+    function validateVoteOption(uint256 campaignID, uint256 option) internal view {
```
manhlx3006, line 773: I already had this in my cleanup dao PR: https://github.com/KyberNetwork/smart-contracts/pull/972/files
Should we close this PR?

contracts/sol6/Dao/KyberDAO.sol
```diff
@@ -682,7 +717,7 @@ contract KyberDAO is IKyberDAO, EpochUtils, ReentrancyGuard, Utils5, DaoOperator
         uint256 cInPrecision,
         uint256 tInPrecision,
         uint256[] memory options
-    ) public view returns (bool) {
+    ) internal view returns (bool) {
```
ilanDoron, line 720: so it shouldn't return bool no more.
since it reverts on failure.

contracts/sol6/Dao/KyberDAO.sol
ilanDoron: I want to squash this PR with other PR for removing pattern of require true.
And both will reference a specific comment by chain security.

contracts/sol6/Dao/KyberDAO.sol
ilanDoron: yes
lets make it internal

PR #971, opened 2020-05-26
Update campaign validation logics in KyberDao, add more get functions in KyberStaking

manhlx3006:
This PR does following changes:
- KyberDao: Move all validations to validateCampaignParams when submit new campaign
- KyberDao: Avoid incompatible startTimestamp and startEpoch in validateCampaignParams function
- KyberDao: Make sure cInPrecision < 2^128 and tInPrecision < 2^128
- KyberStaking: Add functions to get all latest + per epoch data for a staker
- KyberStaking: Update function name for getting raw staker per epoch data


manhlx3006:
Close PR, avoid using audit in branch name

PR #973, opened 2020-05-26
Update campaign validation logics in KyberDao, add more get functions in KyberStaking

manhlx3006:
This PR does following changes:
- KyberDao: Move all validations to validateCampaignParams when submit new campaign
- KyberDao: Avoid incompatible startTimestamp and startEpoch in validateCampaignParams function
- KyberDao: Make sure cInPrecision < 2^128 and tInPrecision < 2^128
- KyberStaking: Add functions to get all latest + per epoch data for a staker
- KyberStaking: Update function name for getting raw staker per epoch data
- KyberStaking: Change delegatedAddress -> representative

### DIFF
--- a/contracts/sol6/Dao/IKyberStaking.sol
+++ b/contracts/sol6/Dao/IKyberStaking.sol
@@ -6,7 +6,7 @@ import "./IEpochUtils.sol";
 interface IKyberStaking is IEpochUtils {
     event Delegated(
         address indexed staker,
-        address indexed delegatedAddress,
+        address indexed representative,
         uint256 indexed epoch,
         bool isDelegated
     );
@@ -16,9 +16,9 @@ interface IKyberStaking is IEpochUtils {
     function initAndReturnStakerDataForCurrentEpoch(address staker)
         external
         returns (
-            uint256 _stake,
-            uint256 _delegatedStake,
-            address _delegatedAddress
+            uint256 stake,
+            uint256 delegatedStake,
+            address representative
         );
 
     function deposit(uint256 amount) external;
@@ -37,8 +37,8 @@ interface IKyberStaking is IEpochUtils {
         external
         view
         returns (
-            uint256 _stake,
-            uint256 _delegatedStake,
-            address _delegatedAddress
+            uint256 stake,
+            uint256 delegatedStake,
+            address representative
         );
 }
--- a/contracts/sol6/Dao/IKyberStaking.sol
+++ b/contracts/sol6/Dao/IKyberStaking.sol
@@ -27,7 +27,13 @@ interface IKyberStaking is IEpochUtils {
 
     function withdraw(uint256 amount) external;
 
-    function getStakerDataForPastEpoch(address staker, uint256 epoch)
+    /**
+     * @notice return raw data of a staker for an epoch
+     *         WARN: should be used only for initialized data
+     *          if data has not been initialized, it will return all 0
+     *          pool master shouldn't use this function to compute/distribute rewards of pool members
+     */
+    function getStakerRawData(address staker, uint256 epoch)
         external
         view
         returns (
--- a/contracts/sol6/Dao/KyberStaking.sol
+++ b/contracts/sol6/Dao/KyberStaking.sol
@@ -16,7 +16,7 @@ contract KyberStaking is IKyberStaking, EpochUtils, ReentrancyGuard {
     struct StakerData {
         uint256 stake;
         uint256 delegatedStake;
-        address delegatedAddress;
+        address representative;
     }
 
     IERC20 public kncToken;
@@ -24,7 +24,7 @@ contract KyberStaking is IKyberStaking, EpochUtils, ReentrancyGuard {
 
     // staker data per epoch
     mapping(uint256 => mapping(address => StakerData)) internal stakerPerEpochData;
-    // latest data of a staker, including stake, delegated stake, delegated address
+    // latest data of a staker, including stake, delegated stake, representative
     mapping(address => StakerData) internal stakerLatestData;
     // true/false: if we have inited data at an epoch for a staker
     mapping(uint256 => mapping(address => bool)) internal hasInited;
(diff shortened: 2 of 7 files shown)


contracts/sol6/Dao/IKyberStaking.sol
```diff
@@ -27,7 +27,7 @@ interface IKyberStaking is IEpochUtils {
 
     function withdraw(uint256 amount) external;
 
-    function getStakerDataForPastEpoch(address staker, uint256 epoch)
+    function getStakerRawData(address staker, uint256 epoch)
```
ilanDoron, line 30: better add 1 comment here 
that this function returns valid values only for a staker that didn't delegate

contracts/sol6/Dao/KyberStaking.sol
```diff
@@ -201,11 +200,13 @@ contract KyberStaking is IKyberStaking, EpochUtils, ReentrancyGuard {
     }
 
     /**
+     * @notice return raw data of a staker for a specific epoch
+     *         WARN: data may not be accurate if data has not been inited
```
ilanDoron, line 204: WARNING: should be used only for init data. So use only for a staker that didn't delegate.

so its clear its not for pool operator.

and have same warning in interface.

contracts/sol6/Dao/KyberStaking.sol
```diff
@@ -303,6 +338,21 @@ contract KyberStaking is IKyberStaking, EpochUtils, ReentrancyGuard {
...
+        )
+    {
+        stake = stakerLatestData[staker].stake;
+        delegatedStake = stakerLatestData[staker].delegatedStake;
+        delegatedAddress = stakerLatestData[staker].delegatedAddress == address(0)
+                ? staker
+                : stakerLatestData[staker].delegatedAddress;
+    }
+
     // prettier-ignore
```
ilanDoron, line 356: BTW
Do we still need prettier ignore here?

Some of those were left when we used shorter lines (80)

test/sol6/kyberDao.js
```diff
@@ -1042,7 +1042,7 @@ contract('KyberDAO', function(accounts) {
           1, startBlock + 2, startBlock + 2 + minCampPeriod,
           minPercentageInPrecision, cInPrecision, tInPrecision, [1, 2, 3], '0x', {from: daoOperator}
         ),
-        "newCampaign: already had network fee for this epoch"
+        "validateParams: already had network fee for this epoch"
```
ilanDoron, line 1045: I think can write
already had network fee campaign for ...

contracts/sol6/Dao/IKyberStaking.sol
manhlx3006: Not exactly, only return valid value if staker has done at least 1 action at previous or that epoch

contracts/sol6/Dao/KyberStaking.sol
manhlx3006: Removed

test/sol6/kyberDao.js
manhlx3006: Updated

contracts/sol6/Dao/IKyberStaking.sol
ilanDoron: but its 99% that if he delegated.
he has done no action.

so pool operator shouldn't be using this by design...

manhlx3006: yeah, if he didn't delegate, he should have voted for some campaigns

manhlx3006: Added more comment there, please help to re-check

ilanDoron: OK
I think should its crystal clear now - for pool operators

contracts/sol6/Dao/KyberStaking.sol
```diff
@@ -288,6 +286,43 @@ contract KyberStaking is IKyberStaking, EpochUtils, ReentrancyGuard {
...
     }
 
+    /**
+     * @notice return combine data (stake, delegatedStake, delegatedAddress) of a staker
+     * @dev allow to get staker data up to current epoch + 1
+     */
+    function getStakerData(address staker, uint256 epoch)
+        external view
+        returns (
+            uint256 stake,
```
ilanDoron, line 296: here should later update names.
if we find better ones.

Seems to me we should

users will always make every possible mistake
so we should minimize mistake options...
:)

contracts/sol6/Dao/mock/MockStakingContract.sol
```diff
@@ -27,11 +27,11 @@ contract MockStakingContract is KyberStaking {
         return stakerPerEpochData[epoch][staker].delegatedStake;
     }
 
-    function getDelegatedAddressValue(address staker, uint256 epoch)
+    function getRepresentativeValue(address staker, uint256 epoch)
```
ilanDoron, line 30: why do we need this getter in mock?

manhlx3006: Using this to get raw data of representative without looping back, so can test correct data updated for each epoch

ilanDoron: OK

contracts/sol6/Dao/IKyberStaking.sol
```diff
@@ -18,7 +18,7 @@ interface IKyberStaking is IEpochUtils {
         returns (
             uint256 _stake,
```
Anyhowclick, line 19: Is there a reason why the return params for this function have underscore prefix? Otherwise, seems to make more sense to align it with the other functions.

Anyhowclick, line 19: Is there a reason why the return params for this function have underscore prefix? Otherwise, seems to make more sense to align it with the other functions.

manhlx3006: Updated, last time we have stake, delegatedStake, delegatedAddress as a global variables, so have to use `_` as prefix

Anyhowclick: ohhhhhhhhhhhh

PR #975, opened 2020-05-26
Validate that token is listed for reserve in KyberHintHandler

ayobuenavista:
Other improvements: MaskOut trade type with empty reserveId list is valid

### DIFF
--- a/contracts/sol6/KyberHintHandler.sol
+++ b/contracts/sol6/KyberHintHandler.sol
@@ -21,13 +21,14 @@ import "./IKyberHint.sol";
  */
 abstract contract KyberHintHandler is IKyberHint, Utils5 {
     /// @notice Parses the hint for a token to ether trade
+    /// @param tokenSrc source token to trade
     /// @param hint The ABI encoded hint, built using the build*Hint functions
     /// @return tokenToEthType Decoded hint type
     /// @return tokenToEthReserveIds Decoded reserve IDs
     /// @return tokenToEthAddresses Reserve addresses corresponding to reserve IDs
     /// @return tokenToEthSplits Decoded splits
-    function parseTokenToEthHint(bytes calldata hint)
-        external
+    function parseTokenToEthHint(IERC20 tokenSrc, bytes memory hint)
+        public
         view
         override
         returns (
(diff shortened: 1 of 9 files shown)


contracts/sol6/KyberMatchingEngine.sol
```diff
@@ -204,6 +209,17 @@ contract KyberMatchingEngine is KyberHintHandler, IKyberMatchingEngine, Withdraw
         (reserveAddress, , , ,) = kyberStorage.getReserveDetailsById(reserveId);
     }
 
+    function areAllReservesListed(
+        bytes32[] memory reserveIds,
+        IERC20 src,
+        IERC20 dest
+    ) internal override view returns (bool) {
```
ilanDoron, line 216: please use boolean values

returns(bool result)
and don't declare another one.

contracts/sol6/KyberMatchingEngine.sol
```diff
@@ -94,14 +94,19 @@ contract KyberMatchingEngine is KyberHintHandler, IKyberMatchingEngine, Withdraw
...
             }
             if (dest == ETH_TOKEN_ADDRESS) {
                 (unpackedHint, ) = unpackT2THint(hint);
-                (tradeType, reserveIds, splitValuesBps, error) = parseHint(unpackedHint);
+                (tradeType, reserveIds, splitValuesBps, error) = parseHint(src, unpackedHint, true);
             }
         } else {
-            (tradeType, reserveIds, splitValuesBps, error) = parseHint(hint);
+            if (src == ETH_TOKEN_ADDRESS) {
+                (tradeType, reserveIds, splitValuesBps, error) = parseHint(dest, hint, false);
```
ilanDoron, line 105: we decided network should do this validation.
And it shouldn't trust matching engine

seems to me now the validation will be done twice for trade and get rate
right?

contracts/sol6/KyberMatchingEngine.sol
ayobuenavista: Fixed

contracts/sol6/KyberMatchingEngine.sol
ayobuenavista: Changed to external functions only.

contracts/sol6/KyberHintHandler.sol
```diff
@@ -259,7 +274,7 @@ abstract contract KyberHintHandler is IKyberHint, Utils5 {
         hint = abi.encode(t2eHint, e2tHint);
     }
 
-    /// @notice Parses or decodes the token to ether or ether to token bytes hint
+    /// @notice Parses or decodes the Token to ETH or ETH to Token bytes hint
```
Anyhowclick, line 277: err, why change this ah?

Anyhowclick: I'll approve first, either Anton fix, or we fix in cleanup

PR #984, opened 2020-05-29
IKyberNetworkProxy will inehrit ISimpleKyberProxy and enable easy usa…

ilanDoron:
…ge of interface files. Currently a user that wants to use the simple API must use both interfaces.


ilanDoron:
moved to another PR

PR #994, opened 2020-06-03
Add missing functions in interface files (#988)

ilanDoron:


contracts/sol6/Dao/IKyberStaking.sol
```diff
@@ -27,6 +27,40 @@ interface IKyberStaking is IEpochUtils {
...
+    function getStakerData(address staker, uint256 epoch)
+        external
+        view
+        returns (
+            uint256 stake,
+            uint256 delegatedStake,
+            address representative
+        );
+
+    function getStake(address staker, uint256 epoch) external view returns (uint256);
```
ilanDoron, line 43: I am thinking the 3 separate functions should not be here.

its for "advanced" users

if they want it.
they should write their own interface

Anyhowclick: I don't see any downside having it... Developers are lazy ppl, asking them to do more work isn't nice =)

PR #996, opened 2020-06-04
Remove FeeHandler from KyberDao, move claim staker reward to FeeHandler

manhlx3006:
- Remove FeeHandler and claimReward from KyberDao
- Move claim staker reward to FeeHandler, and pull staker's reward percentage from KyberDao
- Separate get staker's reward percentage into 2 functions: one returns data for past epochs only, the other one returns data for current epoch
- Add token address into claim events in FeeHandler


contracts/sol6/IKyberFeeHandler.sol
```diff
@@ -1,11 +1,13 @@
...
 
 interface IKyberFeeHandler {
-    event RewardPaid(address indexed staker, uint256 indexed epoch, uint256 amountWei);
-    event RebatePaid(address indexed rebateWallet, uint256 amountWei);
-    event PlatformFeePaid(address indexed platformWallet, uint256 amountWei);
-    event KncBurned(uint256 kncTWei, uint256 amountWei);
+    event RewardPaid(address indexed staker, uint256 indexed epoch, IERC20 indexed token, uint256 amount);
+    event RebatePaid(address indexed rebateWallet, IERC20 indexed token, uint256 amount);
+    event PlatformFeePaid(address indexed platformWallet, IERC20 indexed token, uint256 amount);
+    event KncBurned(uint256 kncTWei, IERC20 indexed token, uint256 amount);
```
ilanDoron, line 10: maybe here amount should be tokenAmount

contracts/sol6/IKyberFeeHandler.sol
```diff
@@ -1,11 +1,13 @@
...
 
     function handleFees(
```
Anyhowclick, line 12: Maybe `platformFeeWei` -> `platformFeeTwei`?

Anyhowclick: Also, `uint` -> `uint256` for the burnAmountWei

Anyhowclick: If we do `platformFeeWei -> platformFeeTwei`, then should do for all the other params as well.. like `rewardWei`, `rebateWei`, `burnAmountWei`....

manhlx3006: The events are designed for fee in ETH + tokens, but current FeeHandler is designed only for fee in ETH, I think the params in current FeeHandler should be still `wei` amount.

Anyhowclick: yeah ok

contracts/sol6/IKyberFeeHandler.sol
manhlx3006: Changed to amountTwei

PR #997, opened 2020-06-04
Fix solhint setup  in CI. Fix linting errors.

ilanDoron:


ilanDoron:
Can't merge at this stage.
will add this fix later.

PR #999, opened 2020-06-05
Add brrWei to handleFees for expected tokenFeeHandlers

Anyhowclick:


contracts/sol6/Dao/KyberFeeHandler.sol
```diff
@@ -154,25 +160,28 @@ contract KyberFeeHandler is IKyberFeeHandler, Utils5, DaoOperator, ReentrancyGua
...
+        IERC20 token,
         address[] calldata rebateWallets,
         uint256[] calldata rebateBpsPerWallet,
         address platformWallet,
-        uint256 platformFeeWei
+        uint256 platformFeeWei,
+        uint256 feeBRRWei
     ) external payable override onlyKyberNetwork nonReentrant {
-        require(msg.value >= platformFeeWei, "msg.value low");
+        token; // not used in this feeHandler
```
ilanDoron, line 177: can assert to ETH

contracts/sol6/Dao/KyberFeeHandler.sol
```diff
@@ -188,29 +197,28 @@ contract KyberFeeHandler is IKyberFeeHandler, Utils5, DaoOperator, ReentrancyGua
             return;
         }
 
-        uint256 rebateWei;
-        uint256 rewardWei;
+        BRRWei memory brrData;
```
ilanDoron, line 200: maybe BrrAmounts.

Data is like the global one that has epoch as well.

contracts/sol6/KyberNetwork.sol
```diff
@@ -1026,9 +1028,9 @@ contract KyberNetwork is WithdrawableNoModifiers, Utils5, IKyberNetwork, Reentra
         require(input.src != input.dest, "src = dest");
 
         if (input.src == ETH_TOKEN_ADDRESS) {
-            require(msg.value == input.srcAmount, "bad eth qty");
```
manhlx3006, line 1029: why do we remove revert messages here?

contracts/sol6/IKyberFeeHandler.sol
```diff
@@ -10,10 +10,12 @@ interface IKyberFeeHandler {
     event KncBurned(uint256 kncTWei, IERC20 indexed token, uint256 amountTwei);
 
     function handleFees(
+        IERC20 token,
```
manhlx3006, line 13: should we use `Twei` here as well?

contracts/sol6/KyberNetwork.sol
Anyhowclick: I think the main reason was that it the proxy will also catch this case...? But it probably will reach this line first.

The other reason is so that we increase the solc runs.

Can put the msg back.

contracts/sol6/IKyberFeeHandler.sol
Anyhowclick: `platformFeeTwei` and `feeBRRTwei`? Mmmm...

manhlx3006: The interface APIs are designed for both eth + token as fee right?

ilanDoron: for Eth we need wei
for token we need twei.
can't have them both.

we could call one:
platformFee
the other
networkFee

and in comment say it is wei for eth, twei for token.

manhlx3006: Then for event we also need to change amount name?

contracts/sol6/KyberNetwork.sol
ilanDoron: Reason is it shouldn't happen.
and if it does it should be caught on proxy level and it will issue a message (we should fix proxy message...)

so enough to have require without message for cases that will revert on proxy level.

But @Anyhowclick why u say it will reach this line first?

manhlx3006: Proxy doesn't check if srcAmount == msg.value if src == ETH, that's why it will be reverted here in Network first

ilanDoron: Right
I fixed that now in another PR
we better have it on proxy level.

contracts/sol6/IKyberFeeHandler.sol
ilanDoron: not sure.

contracts/sol6/KyberNetwork.sol
Anyhowclick: Ok, I'm reverting any changes made to revert msgs.

contracts/sol6/IKyberFeeHandler.sol
Anyhowclick: Ok, will change.

PR #1000, opened 2020-06-05
add test for missing event

ductm54:


test/sol6/kyberDao.js
```diff
@@ -409,6 +410,17 @@ contract('KyberDao', function(accounts) {
...
         link: link
       });
+      let eventLogs;
+      for (let i = 0; i < txResult.logs.length; i++) {
+          if (txResult.logs[i].event == 'NewCampaignCreated') {
+              eventLogs = txResult.logs[i];
+              break;
+          }
+      }
+      Helper.assertEqual(eventLogs.args.options.length, options.length);
```
ilanDoron, line 420: can use maybe the array equal API Quang added?

ductm54: updated

PR #1005, opened 2020-06-12
Fix deployment script

Anyhowclick:
- Uses compiled artifacts in artifacts folder instead of re-compiling
- Update to be compatible with latest Katalyst changes
- Added redeployment scripts and checks for network and proxy


web3deployment/KatalystDeploy/katalystNetworkDeployer.js
```diff
@@ -236,9 +249,8 @@ async function main() {
 
   chainId = chainId || await web3.eth.net.getId()
   console.log('chainId', chainId);
-  console.log('starting compilation');
-  output = await require("../compileContracts.js").compileContracts("sol5");
-  console.log("finished compilation");
+  console.log('retrieving artifacts...');
+  output = await require("../retrieveArtifacts.js").retrieveArtifacts();
```
ilanDoron, line 253: this means have to compile before starting?

what if compiled version is wrong?
or not the expected one?

do handle this situation?

web3deployment/retrieveArtifacts.js
```diff
@@ -0,0 +1,37 @@
...
+const artifactsPath = path.join(__dirname, "../artifacts/");
+const execSync = require('child_process').execSync;
+
+module.exports.retrieveArtifacts = main;
+async function main() {
+  if (fs.existsSync(artifactsPath)) {
+    let output = await packageArtifacts();
+    return output;
+  } else {
+    compileArtifacts();
```
ilanDoron, line 14: Nice.
So this is much simpler than we had.
Now I understand your comment about it.

I think correct function name is compile contracts.

artifacts is result. 
no?

web3deployment/retrieveArtifacts.js
```diff
@@ -0,0 +1,37 @@
...
+  files.forEach(file => {
+    content = JSON.parse(fs.readFileSync(path.join(artifactsPath, file)));
+    result[content.contractName] = content;
+  })
+  return result;
+}
+
+
+function compileArtifacts() {
+  console.log("Artifacts not found. Compiling contracts...");
```
ilanDoron, line 31: what if artifacts are found the not updated?
Can u check?

web3deployment/retrieveArtifacts.js
```diff
@@ -0,0 +1,38 @@
...
+const readdir = util.promisify(fs.readdir);
+const path = require("path");
+const artifactsPath = path.join(__dirname, "../artifacts/");
+const buidlerConfigSol5 = path.join(__dirname, "../buidlerConfigSol5.js");
+const buidlerConfigSol4 = path.join(__dirname, "../buidlerConfigSol4.js");
+const execSync = require('child_process').execSync;
+
+module.exports.retrieveArtifacts = main;
+async function main(skipCompilation) {
+  if (!skipCompilation) {
```
ilanDoron, line 12: where is this defined?

web3deployment/retrieveArtifacts.js
```diff
@@ -0,0 +1,38 @@
...
+    compileContracts();
+  }
+  let output = await packageArtifacts();
+  console.log(output)
```
ayobuenavista, line 16: Might want to remove this since it's only good for debugging. Fills the screen with a ton of output.

ilanDoron: @Anyhowclick 
we have the option to log to screen only errors?

like we did when using compile

web3deployment/retrieveArtifacts.js
Anyhowclick: For now, no, but if you've already done the compilation before, can pass in a `true` value. Just a nice to have

web3deployment/retrieveArtifacts.js
Anyhowclick: ohh right, will remove it. 
Since it's running `npx buidler compile`, any compilation errors should automatically be logged

PR #1013, opened 2020-06-21
Added solhint config

ayobuenavista:


.solhintignore
```diff
@@ -1,4 +1,8 @@
...
-contracts/sol6/utils/zeppelin
-contracts/sol5/bridges
-
+contracts/sol4
+contracts/sol5
+contracts/sol6/Dao
+contracts/sol6/mock
+contracts/sol6/utils
+contracts/sol6/wrappers
+contracts/sol6/bridges
```
ilanDoron, line 7: please remove bridges.
WE don't have this folder.
And new files should be affected.

ayobuenavista: Regarding solhint only affecting new files, it is a bit tricky since we're using wildcards. For this to work, we need list at each file in solhintignore. Otherwise, as long as new files are added in new folders, e.g. "sol6/reserves", then it will work and solhint will scan it.

ilanDoron: yes
u r right.
so for now it should handle new files in new folders.
later we can extend.

PR #1018, opened 2020-06-22
Deduplication PR for emptyHintFix PR

ayobuenavista:


ayobuenavista:
Made a PR to emptyHintFix PR based on discussion with Desmond. He suggested since `splits != 0`, is checked twice, might as well remove the check in the first one. So the `SplitsNotEmptyError` checks for BestOfAll, MaskIn, MaskOut now (previously just MaskIn, MaskOut).

Can just close this if it's not needed, as the previous PR is already valid and working as well.

Anyhowclick:
Ilan said it's too late, so... gonna close it

PR #1050, opened 2020-07-31
[don't pull]add deployer with enhancedStepsReserve

ducquangkstn:


web3deployment/reserveDeploy/fpr/enhancedStepsDeployer.js
```diff
@@ -0,0 +1,76 @@
...
+let wrapper;
+let wrapperAddr = "";
+
+let networkAddr = "0x920b322d4b8bab34fb6233646f5c87f87e79952b";
+let admin= "0xf3D872b9E8d314820dc8E99DAfBe1A3FeEDc27D5";
+let deployer;
+
+async function main() {
+    const accounts = await web3.eth.getAccounts();
+    deployer = accounts[0];
```
ilanDoron, line 20: plan is to send funds to this address?
if so how to return back remaining funds when done?
have private key?
consider saving it locally.

web3deployment/reserveDeploy/fpr/enhancedStepsDeployer.js
```diff
@@ -0,0 +1,76 @@
...
+    }
+}
+
+
+main()
+  .then(() => process.exit(0))
+  .catch(error => {
+    console.error(error);
+    process.exit(1);
+  });
```
ilanDoron, line 76: new line

web3deployment/reserveDeploy/fpr/enhancedStepsDeployer.js
```diff
@@ -0,0 +1,76 @@
...
+let networkAddr = "0x920b322d4b8bab34fb6233646f5c87f87e79952b";
+let admin= "0xf3D872b9E8d314820dc8E99DAfBe1A3FeEDc27D5";
+let deployer;
+
+async function main() {
+    const accounts = await web3.eth.getAccounts();
+    deployer = accounts[0];
+    console.log(`deployer address at ${deployer}`);
+
+    gasPrice = new BN(75).mul(new BN(10).pow(new BN(9)))
```
ilanDoron, line 23: maybe use yargs for gas price?

ducquangkstn: if we run the script with buidler, we can not read value from args
but we can read the value from environment variable

web3deployment/reserveDeploy/fpr/enhancedStepsDeployer.js
ducquangkstn: if you set the network is ropsten in the script, accounts[0] will be the address from the privatekey in .env file

PR #1052, opened 2020-08-04
New reserve push. Kyber reserve + conversion rate enhanced steps with higher max_rate.

ilanDoron:
Current ConversionRateEnhancedSteps inherits MAX_RATE from utils contract and uses it in addBps function.

currently our code uses PRECISION * 10^6
but in Katalyst MAX_RATE is PRECISION * 10^7
for using the bigger value. this PR updates MAX_RATE in use for two contracts:
KyberReseve - creates a new contract KyberReserveHighRate
and updating enhanced steps contract, which can be updated since not yet used in production.

code was written to enable max code reuse and minimal code duplication

Wrapper for enhanced conversion rates was modified:

remove call to setQtyStepFunction which can't be used in enhanced conversion rate.
call set imbalance with x and y arrays with different length.


contracts/sol4/mock/ConversionRatesEnhancedOpen.sol
```diff
@@ -0,0 +1,106 @@
...
+    ) public view returns(
+        uint rateWithSteps,
+        uint rateWithoutSteps,
+        int extraBpsYdata
+    ) {
+        // get rate update block
+        bytes32 compactData = tokenRatesCompactData[tokenData[token].compactDataArrayIndex];
+
+        uint updateRateBlock = getLast4Bytes(compactData);
+        if (currentBlockNumber >= updateRateBlock + validRateDurationInBlocks) return (0, 0, 0); // rate is expired
```
manhlx3006, line 37: it does check for valid rate duration here, but comment mentions it doesn't check

ilanDoron: will remove this check
also taking block.number instead of user input block number.
Not sure if we ever use this contract again. but better its easy to use.

Anyhowclick (approved):

manhlx3006 (approved):

ducquangkstn (approved):

PR #1054, opened 2020-08-05
travis test all file

ductm54:


.travis/test-part.sh
```diff
@@ -6,8 +6,8 @@ set -euxo pipefail
 readonly test_part=${TEST_PART:-}
 
 case "$test_part" in
-Sol6)
-    npx buidler test --no-compile --config buidlerConfigSol6.js
+Test)
```
ilanDoron, line 9: in test file can maybe leave both options
so later we can configure as we want for yml file

ilanDoron: actually seems more clear to call it All

and leave sol6 as well.

.travis.yml
```diff
@@ -13,6 +13,8 @@ env:
 
 jobs:
   include:
+    - env:
+        - TEST_PART: All
```
ilanDoron, line 17: when written like this it will test both All and Sol6?

buidler.config.js
```diff
@@ -134,7 +134,8 @@ module.exports = {
             "0x28d1bfbbafe9d1d4f5a11c3c16ab6bf9084de48d99fbac4058bdfa3c80b2908f",
           balance: "100000000000000000000000000000000"
         }
-      ]
+      ],
+      allowUnlimitedContractSize: true
```
ilanDoron, line 138: why do we need this part?

buidlerConfigSol6.js
```diff
@@ -135,7 +135,8 @@ module.exports = {
             "0x28d1bfbbafe9d1d4f5a11c3c16ab6bf9084de48d99fbac4058bdfa3c80b2908f",
           balance: "100000000000000000000000000000000"
         }
-      ]
+      ],
+      allowUnlimitedContractSize: true
```
ilanDoron, line 139: and if we need it.
is it required both sol4 and sol6?

test/sol4/kyberProxyV1.js
```diff
@@ -1571,19 +1571,16 @@ contract('KyberProxyV1', function(accounts) {
         });
 
         it("verify sell with malicious network2 reverts when using any min rate (0).", async function () {
-            let amountTwei = 1123;
+            let amountTwei = new BN(1000);
 
-            //set steal amount to 1 wei
-            let myFee = 1;
+            //set steal amount to 100 wei
+            let myFee = 100;
```
ilanDoron, line 1577: 1 wei doesn't work anymore?

test/sol6/kyberNetwork.js
```diff
@@ -1060,7 +1060,8 @@ contract('KyberNetwork', function(accounts) {
                 });
 
                 // 2 types of trade that specific which reserve will be choosen
```
ilanDoron, line 1062: if we are here.
can you fix the comment

seems it should have
specific --> specify

choosen --> chosen

buidler.config.js
Anyhowclick: The malicious network contracts exceeds the max bytecode size. Without this, tests will fail cos the contracts can't be deployed (out of gas txns).

buidlerConfigSol6.js
Anyhowclick: Only sol6, cos it's only the malicious network contracts that exceed.

PR #1056, opened 2020-08-06
Updated solhint ignored files and solhint rules

ayobuenavista:


.solhintignore
```diff
@@ -1,7 +1,127 @@
-contracts/sol4
-contracts/sol5
-contracts/sol6/Dao
-contracts/sol6/mock
-contracts/sol6/utils
-contracts/sol6/wrappers
-contracts/sol6/*.sol
+contracts/sol4/ConversionRatesInterface.sol
```
ilanDoron, line 1: please organise a bit so its easier to find a file.

example so first all files that are
sol4/*.*
sol4/mock
sol4/wrappers

and add some headlines.
same as gitignore file.

ayobuenavista: More organized now

PR #1059, opened 2020-08-21
Add KyberFprReserveV2 supporting WETH as quote

manhlx3006:
KyberFprReserve version 2
- Allow Reserve to work with either WETH (by specifying address to hold WETH) or ETH
- Allow Reserve to set maxGasPriceWei to trade with, this value is set by only Operator

### DIFF
--- a/contracts/sol6/reserves/IConversionRates.sol
+++ b/contracts/sol6/reserves/IConversionRates.sol
@@ -1,0 +1,22 @@
+pragma solidity 0.6.6;
+
+
+import "../IERC20.sol";
+
+
+interface IConversionRates {
+
+    function recordImbalance(
+        IERC20 token,
+        int buyAmount,
+        uint256 rateUpdateBlock,
+        uint256 currentBlock
+    ) external;
+
+    function getRate(
+        IERC20 token,
+        uint256 currentBlockNumber,
+        bool buy,
+        uint256 qty
+    ) external view returns(uint256);
+}
--- a/contracts/sol6/reserves/IWeth.sol
+++ b/contracts/sol6/reserves/IWeth.sol
@@ -1,0 +1,10 @@
+pragma solidity 0.6.6;
+
+import "../IERC20.sol";
+
+
+interface IWeth is IERC20 {
+    function deposit() external payable;
+    function withdraw(uint256 wad) external;
+}
+
--- a/contracts/sol6/reserves/IWeth.sol
+++ b/contracts/sol6/reserves/IWeth.sol
@@ -7,4 +7,3 @@ interface IWeth is IERC20 {
     function deposit() external payable;
     function withdraw(uint256 wad) external;
 }
-
--- a/contracts/sol6/reserves/KyberFprReserveV2.sol
+++ b/contracts/sol6/reserves/KyberFprReserveV2.sol
@@ -88,6 +88,7 @@ contract KyberFprReserveV2 is IKyberReserve, Utils5, Withdrawable3 {
     {
         require(tradeEnabled, "trade not enable");
         require(msg.sender == kyberNetwork, "wrong sender");
+        require(tx.gasprice <= maxGasPriceWei, "gas price too high");
 
         doTrade(srcToken, srcAmount, destToken, destAddress, conversionRate, validate);
 
--- a/contracts/sol6/reserves/KyberFprReserveV2.sol
+++ b/contracts/sol6/reserves/KyberFprReserveV2.sol
@@ -86,16 +86,14 @@ contract KyberFprReserveV2 is IKyberReserve, Utils5, Withdrawable3 {
         return true;
     }
 
-    function enableTrade() external onlyAdmin returns (bool) {
+    function enableTrade() external onlyAdmin {
         tradeEnabled = true;
         emit TradeEnabled(true);
-        return true;
     }
 
-    function disableTrade() external onlyAlerter returns (bool) {
+    function disableTrade() external onlyAlerter {
         tradeEnabled = false;
         emit TradeEnabled(false);
-        return true;
     }
 
     function setMaxGasPrice(uint256 newMaxGasPrice) external onlyOperator {
--- a/contracts/sol6/reserves/KyberFprReserveV2.sol
+++ b/contracts/sol6/reserves/KyberFprReserveV2.sol
@@ -11,6 +11,9 @@ import "../utils/Withdrawable3.sol";
 import "../utils/zeppelin/SafeERC20.sol";
 
 
+/// @title KyberFrpReserve version 2
+/// Allow Reserve to work with both WETH and ETH by specifying address to hold WETH
+/// Allow Reserve to set maxGasPriceWei to trade with
 contract KyberFprReserveV2 is IKyberReserve, Utils5, Withdrawable3 {
     using SafeERC20 for IERC20;
     using SafeMath for uint256;
@@ -20,7 +23,7 @@ contract KyberFprReserveV2 is IKyberReserve, Utils5, Withdrawable3 {
 
     address public kyberNetwork;
     bool public tradeEnabled;
-    uint256 public maxGasPrice;
+    uint256 public maxGasPriceWei;
 
     IConversionRates public conversionRatesContract;
     ISanityRates public sanityRatesContract;
@@ -51,19 +54,19 @@ contract KyberFprReserveV2 is IKyberReserve, Utils5, Withdrawable3 {
         address _kyberNetwork,
         IConversionRates _ratesContract,
         IWeth _weth,
-        uint256 _maxGasPrice,
+        uint256 _maxGasPriceWei,
         address _admin
     )
         Withdrawable3(_admin) public
     {
-        require(_ratesContract != IConversionRates(0), "ratesContract 0");
         require(_kyberNetwork != address(0), "kyberNetwork 0");
+        require(_ratesContract != IConversionRates(0), "ratesContract 0");
         require(_weth != IWeth(0), "weth 0");
         kyberNetwork = _kyberNetwork;
         conversionRatesContract = _ratesContract;
         weth = _weth;
         tradeEnabled = true;
-        maxGasPrice = _maxGasPrice;
+        maxGasPriceWei = _maxGasPriceWei;
     }
 
     receive() external payable {
@@ -92,26 +95,20 @@ contract KyberFprReserveV2 is IKyberReserve, Utils5, Withdrawable3 {
     }
 
     function enableTrade() onlyAdmin external returns(bool) {
-        if (!tradeEnabled) {
-            tradeEnabled = true;
-            emit TradeEnabled(true);
-        }
+        tradeEnabled = true;
+        emit TradeEnabled(true);
         return true;
     }
 
     function disableTrade() onlyAlerter external returns(bool) {
-        if (tradeEnabled) {
-            tradeEnabled = false;
-            emit TradeEnabled(false);
-        }
+        tradeEnabled = false;
+        emit TradeEnabled(false);
         return true;
     }
 
-    function setMaxGasPrice(uint256 newMaxGasPrice) onlyAdmin external {
-        if (maxGasPrice != newMaxGasPrice) {
-            maxGasPrice = newMaxGasPrice;
-            emit MaxGasPriceUpdated(newMaxGasPrice);
-        }
+    function setMaxGasPrice(uint256 newMaxGasPrice) onlyOperator external {
+        maxGasPriceWei = newMaxGasPrice;
+        emit MaxGasPriceUpdated(newMaxGasPrice);
     }
 
     function approveWithdrawAddress(IERC20 token, address addr, bool approve) onlyAdmin external {
@@ -122,11 +119,9 @@ contract KyberFprReserveV2 is IKyberReserve, Utils5, Withdrawable3 {
 
     /// @dev allow set tokenWallet[token] back to 0x0 address
     function setTokenWallet(IERC20 token, address wallet) onlyAdmin external {
-        if (tokenWallet[address(token)] != wallet) {
-            tokenWallet[address(token)] = wallet;
-            emit NewTokenWallet(token, wallet);
-        }
+        tokenWallet[address(token)] = wallet;
         setDecimals(token);
+        emit NewTokenWallet(token, wallet);
     }
 
     /// @dev withdraw amount of token to an approved destination
@@ -190,7 +185,8 @@ contract KyberFprReserveV2 is IKyberReserve, Utils5, Withdrawable3 {
         returns(uint256)
     {
         if (!tradeEnabled) return 0;
-        if (tx.gasprice > maxGasPrice) return 0;
+        if (tx.gasprice > maxGasPriceWei) return 0;
+        if (srcQty == 0) return 0;
 
         IERC20 token;
         bool isBuy;
@@ -216,6 +212,10 @@ contract KyberFprReserveV2 is IKyberReserve, Utils5, Withdrawable3 {
         }
 
         return rate;
+    }
+
+    function isAddressApprovedForWithdrawal(IERC20 token, address addr) external view returns(bool) {
+        return approvedWithdrawAddresses[keccak256(abi.encodePacked(address(token), addr))];
     }
 
     function getBalance(IERC20 token) public view returns(uint256) {
--- a/contracts/sol6/reserves/KyberFprReserveV2.sol
+++ b/contracts/sol6/reserves/KyberFprReserveV2.sol
@@ -1,4 +1,3 @@
-
 pragma solidity 0.6.6;
 
 import "./IConversionRates.sol";
@@ -10,7 +9,6 @@ import "../utils/Utils5.sol";
 import "../utils/Withdrawable3.sol";
 import "../utils/zeppelin/SafeERC20.sol";
 
-
 /// @title KyberFprReserve version 2
 /// Allow Reserve to work with both WETH by specifying address to hold WETH or ETH
 /// Allow Reserve to set maxGasPriceWei to trade with
@@ -18,8 +16,8 @@ contract KyberFprReserveV2 is IKyberReserve, Utils5, Withdrawable3 {
     using SafeERC20 for IERC20;
     using SafeMath for uint256;
 
-    mapping(bytes32=>bool) public approvedWithdrawAddresses; // sha3(token,address)=>bool
-    mapping(address=>address) public tokenWallet;
+    mapping(bytes32 => bool) public approvedWithdrawAddresses; // sha3(token,address)=>bool
+    mapping(address => address) public tokenWallet;
 
     address public kyberNetwork;
     bool public tradeEnabled;
@@ -56,9 +54,7 @@ contract KyberFprReserveV2 is IKyberReserve, Utils5, Withdrawable3 {
         IWeth _weth,
         uint256 _maxGasPriceWei,
         address _admin
-    )
-        Withdrawable3(_admin) public
-    {
+    ) public Withdrawable3(_admin) {
         require(_kyberNetwork != address(0), "kyberNetwork 0");
         require(_ratesContract != IConversionRates(0), "ratesContract 0");
         require(_weth != IWeth(0), "weth 0");
@@ -80,12 +76,7 @@ contract KyberFprReserveV2 is IKyberReserve, Utils5, Withdrawable3 {
         address payable destAddress,
         uint256 conversionRate,
         bool validate
-    )
-        override
-        external
-        payable
-        returns(bool)
-    {
+    ) external override payable returns (bool) {
         require(tradeEnabled, "trade not enable");
         require(msg.sender == kyberNetwork, "wrong sender");
         require(tx.gasprice <= maxGasPriceWei, "gas price too high");
@@ -95,31 +86,35 @@ contract KyberFprReserveV2 is IKyberReserve, Utils5, Withdrawable3 {
         return true;
     }
 
-    function enableTrade() onlyAdmin external returns(bool) {
+    function enableTrade() external onlyAdmin returns (bool) {
         tradeEnabled = true;
         emit TradeEnabled(true);
         return true;
     }
 
-    function disableTrade() onlyAlerter external returns(bool) {
+    function disableTrade() external onlyAlerter returns (bool) {
         tradeEnabled = false;
         emit TradeEnabled(false);
         return true;
     }
 
-    function setMaxGasPrice(uint256 newMaxGasPrice) onlyOperator external {
+    function setMaxGasPrice(uint256 newMaxGasPrice) external onlyOperator {
         maxGasPriceWei = newMaxGasPrice;
         emit MaxGasPriceUpdated(newMaxGasPrice);
     }
 
-    function approveWithdrawAddress(IERC20 token, address addr, bool approve) onlyAdmin external {
+    function approveWithdrawAddress(
+        IERC20 token,
+        address addr,
+        bool approve
+    ) external onlyAdmin {
         approvedWithdrawAddresses[keccak256(abi.encodePacked(address(token), addr))] = approve;
         setDecimals(token);
         emit WithdrawAddressApproved(token, addr, approve);
     }
 
     /// @dev allow set tokenWallet[token] back to 0x0 address
-    function setTokenWallet(IERC20 token, address wallet) onlyAdmin external {
+    function setTokenWallet(IERC20 token, address wallet) external onlyAdmin {
         tokenWallet[address(token)] = wallet;
         setDecimals(token);
         emit NewTokenWallet(token, wallet);
@@ -130,7 +125,11 @@ contract KyberFprReserveV2 is IKyberReserve, Utils5, Withdrawable3 {
     /// @param token token to withdraw
     /// @param amount amount to withdraw
     /// @param destination address to transfer fund to
-    function withdraw(IERC20 token, uint256 amount, address destination) onlyOperator external {
+    function withdraw(
+        IERC20 token,
+        uint256 amount,
+        address destination
+    ) external onlyOperator {
         require(
             approvedWithdrawAddresses[keccak256(abi.encodePacked(address(token), destination))],
             "destination is not approved"
@@ -156,9 +155,7 @@ contract KyberFprReserveV2 is IKyberReserve, Utils5, Withdrawable3 {
         IConversionRates _conversionRates,
         IWeth _weth,
         IKyberSanity _sanityRates
-    )
-        onlyAdmin external
-    {
+    ) external onlyAdmin {
         require(_kyberNetwork != address(0), "kyberNetwork 0");
         require(_conversionRates != IConversionRates(0), "conversionRates 0");
         require(_weth != IWeth(0), "weth 0");
@@ -181,10 +178,7 @@ contract KyberFprReserveV2 is IKyberReserve, Utils5, Withdrawable3 {
         IERC20 dest,
         uint256 srcQty,
         uint256 blockNumber
-    ) 
-        override external view
-        returns(uint256)
-    {
+    ) external override view returns (uint256) {
         if (!tradeEnabled) return 0;
         if (tx.gasprice > maxGasPriceWei) return 0;
         if (srcQty == 0) return 0;
@@ -215,11 +209,15 @@ contract KyberFprReserveV2 is IKyberReserve, Utils5, Withdrawable3 {
         return rate;
     }
 
-    function isAddressApprovedForWithdrawal(IERC20 token, address addr) external view returns(bool) {
+    function isAddressApprovedForWithdrawal(IERC20 token, address addr)
+        external
+        view
+        returns (bool)
+    {
         return approvedWithdrawAddresses[keccak256(abi.encodePacked(address(token), addr))];
     }
 
-    function getBalance(IERC20 token) public view returns(uint256) {
+    function getBalance(IERC20 token) public view returns (uint256) {
         address wallet = getTokenWallet(token);
         if (token == ETH_TOKEN_ADDRESS) {
             if (wallet == address(this)) {
@@ -239,7 +237,12 @@ contract KyberFprReserveV2 is IKyberReserve, Utils5, Withdrawable3 {
         }
     }
 
-    function getDestQty(IERC20 src, IERC20 dest, uint256 srcQty, uint256 rate) public view returns(uint256) {
+    function getDestQty(
+        IERC20 src,
+        IERC20 dest,
+        uint256 srcQty,
+        uint256 rate
+    ) public view returns (uint256) {
         uint256 dstDecimals = getDecimals(dest);
         uint256 srcDecimals = getDecimals(src);
 
@@ -259,9 +262,7 @@ contract KyberFprReserveV2 is IKyberReserve, Utils5, Withdrawable3 {
         address payable destAddress,
         uint256 conversionRate,
         bool validate
-    )
-        internal
-    {
+    ) internal {
         // can skip validation if done at kyber network level
         if (validate) {
             require(conversionRate > 0, "rate is 0");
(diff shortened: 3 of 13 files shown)


contracts/sol6/reserves/KyberFprReserveV2.sol
```diff
@@ -0,0 +1,349 @@
...
+        }
+
+        uint256 balanceOfWallet = usingToken.balanceOf(wallet);
+        uint256 allowanceOfWallet = usingToken.allowance(wallet, address(this));
+
+        return minOf(balanceOfWallet, allowanceOfWallet);
+    }
+
+    function getDestQty(
+        IERC20 src,
```
ilanDoron, line 253: can use utils5.calcDestAmount()

no?

contracts/sol6/reserves/KyberFprReserveV2.sol
```diff
@@ -0,0 +1,349 @@
...
+    /// @param validate If true, additional validations are applicable
+    function doTrade(
+        IERC20 srcToken,
+        uint256 srcAmount,
+        IERC20 destToken,
+        address payable destAddress,
+        uint256 conversionRate,
+        bool validate
+    ) internal {
+        // can skip validation if done at kyber network level
```
ilanDoron, line 278: lets check this assumption
we are moving toward permission less reserves.
maybe this isn't valid anymore.

on the other hand.
the checks here are not really interesting.
and we know Kyber network will not send wrong value.
and if it does we will probably fix it.


@tranvictor can u comment here please?

contracts/sol6/reserves/KyberFprReserveV2.sol
```diff
@@ -0,0 +1,349 @@
...
+        address destination
+    ) external onlyOperator {
+        require(
+            approvedWithdrawAddresses[keccak256(abi.encodePacked(address(token), destination))],
+            "destination is not approved"
+        );
+
+        if (token == ETH_TOKEN_ADDRESS) {
+            (bool success, ) = destination.call{value: amount}("");
+            require(success, "transfer back eth failed");
```
ilanDoron, line 138: back isn't clear
maybe:
withdraw eth failed

contracts/sol6/reserves/KyberFprReserveV2.sol
```diff
@@ -0,0 +1,349 @@
...
+        if (destToken == ETH_TOKEN_ADDRESS) {
+            if (destTokenWallet != address(this)) {
+                // transfer weth from weth's token wallet to this address
+                // then unwrap weth to eth to this contract address
+                IERC20(weth).safeTransferFrom(destTokenWallet, address(this), destAmount);
+                weth.withdraw(destAmount);
+            }
+            // transfer eth to dest address
+            (bool success, ) = destAddress.call{value: destAmount}("");
+            require(success, "transfer back eth failed");
```
ilanDoron, line 326: back isn't very clear

maybe
transfer eth from reserve to destAddress failed

contracts/sol6/reserves/KyberFprReserveV2.sol
```diff
@@ -0,0 +1,349 @@
...
+            }
+        }
+
+        emit TradeExecute(msg.sender, srcToken, srcAmount, destToken, destAmount, destAddress);
+    }
+
+    /// @dev return wallet that holds the token
+    ///      if token is ETH, check tokenWallet of WETH instead
+    ///      if wallet is 0x0, consider as this reserve address
+    function getTokenWallet(IERC20 token) internal view returns (address wallet) {
```
ilanDoron, line 341: we should consider making this public?
might help reserve manager.

test/sol6/reserves/kyberFprReserveV2.js
```diff
@@ -0,0 +1,2902 @@
+const ConversionRates = artifacts.require("MockEnhancedStepFunctions.sol");
+const TestToken = artifacts.require("TestToken.sol");
+const WethToken = artifacts.require("WethToken.sol");
+const Reserve = artifacts.require("KyberFprReserveV2");
+const MockSanityRates = artifacts.require("MockSanityRates");
+const NoPayableFallback = artifacts.require("NoPayableFallback");
+
```
ilanDoron, line 7: wonder if this reserve should be able to work with apr conversion rate.

I think not, but lets make sure.

test/sol6/reserves/kyberFprReserveV2.js
```diff
@@ -0,0 +1,2902 @@
...
+//imbalance buy steps
+let imbalanceBuyStepX = [-85000, -28000, -15000, 0, 15000, 28000, 45000];
+let imbalanceBuyStepY = [ 1300, 130, 43, 0, 0, -110, -160, -1600];
+
+
+//sell imbalance step
+let imbalanceSellStepX = [-85000, -28000, -10000, 0, 10000, 28000, 45000];
+let imbalanceSellStepY = [-1500, -320, -75, 0, 0, 110, 350, 650];
+
```
ilanDoron, line 76: actually I think we never use positive steps.
wonder if anyone does.

not saying it shouldn't be supported.
I assume u took numbers from previous test file.

test/sol6/reserves/kyberFprReserveV2.js
```diff
@@ -0,0 +1,2902 @@
...
+        operator = accounts[1];
+        network = accounts[2];
+        user1 = accounts[4];
+        user2 = accounts[5];
+        withdrawAddress = accounts[6];
+        sanityRate = accounts[7];
+        alerter = accounts[8];
+        walletForToken = accounts[9];
+        user3 = accounts[8];
+        user4 = accounts[9];
```
ilanDoron, line 98: I think we use 20 accounts by defaults
in the config file.

so can use them if needed. thus don't have to reuse accounts.

test/sol6/reserves/kyberFprReserveV2.js
```diff
@@ -0,0 +1,2902 @@
...
+        // create tokens
+        for (let i = 0; i < numTokens; ++i) {
+            tokenDecimals[i] = new BN(18);
+            token = await TestToken.new("test" + i, "tst" + i, tokenDecimals[i]);
+            tokens[i] = token;
+            tokenAdd[i] = token.address;
+        }
+    });
+
+    const setupConversionRatesContract = async function(needToken) {
```
ilanDoron, line 112: wonder are these functions specific to this test file only
one place we should re use them is when doing gas report file for reserve repository.

@Anyhowclick is planning how build gas report checks for reserve repository
so please comment

ilanDoron (commented):
finished only code review part. and a bit of the tests review.
will review tests in next days.

please note:
1. seems many test values are based on previous tests. Since I wrote them I must say some values are not so representative . at the time I wasn't aware of many stuff. like how to increase eth balance of the accounts. so all is tested with very low eth values.
need to have some test with higher values.

2. make sure we can use it.only for each test.

contracts/sol6/reserves/KyberFprReserveV2.sol
```diff
@@ -0,0 +1,349 @@
...
+import "../utils/zeppelin/SafeERC20.sol";
+
+/// @title KyberFprReserve version 2
+/// Allow Reserve to work with both WETH by specifying address to hold WETH or ETH
+/// Allow Reserve to set maxGasPriceWei to trade with
+contract KyberFprReserveV2 is IKyberReserve, Utils5, Withdrawable3 {
+    using SafeERC20 for IERC20;
+    using SafeMath for uint256;
+
+    mapping(bytes32 => bool) public approvedWithdrawAddresses; // sha3(token,address)=>bool
```
Anyhowclick, line 19: Small thing: Operationally, it's a little bit difficult to do keccak256(token,address) in order to find out whether an address has been whitelisted for approval. Not something that has to be fixed now, cos vv few project teams ask us about it. But it's good if there's a more convenient way.

contracts/sol6/reserves/KyberFprReserveV2.sol
Anyhowclick: +1, I would say it should be public. Not just reserve manager, but for debugging purposes

package.json
```diff
@@ -38,15 +38,15 @@
     "yargs": "11.1.1"
   },
   "devDependencies": {
-    "@nomiclabs/buidler": "1.3.0",
+    "@nomiclabs/buidler": "^1.3.0",
```
Anyhowclick, line 41: remove carat ^

test/sol6/reserves/kyberFprReserveV2.js
Anyhowclick: Would be nice to have a `reservesSetup.js` somewhere. 
My preference is for this file to exist outside the test folder, because I think the gas report file will be outside as well

contracts/sol6/reserves/KyberFprReserveV2.sol
tranvictor: Lets not rely on any trust assumption from the network. Reserves should validate everything before settlement.

ilanDoron: OK

but for rate there is trust relation
we keep that.
right?

since re calculate rate is high on gas.

@tranvictor

tranvictor: No, rate is the most important thing the reserves have to validate otherwise other reserves can trigger a change in the reserve's inventory or some sort in the middle of the process. I think gas cost for rate calculation is not that high.

ilanDoron: OK
lets validate and see gas cost.

What does it mean "can trigger a change in the reserve's inventory or some sort in the middle of the process" - I don't follow this part

contracts/sol6/reserves/KyberFprReserveV2.sol
ilanDoron: there is a getter function below: isAddressApprovedForWithdrawal
is that enough?
or other ideaS?

test/sol6/reserves/kyberFprReserveV2.js
manhlx3006: Ok, will create a file for reserve setup

contracts/sol6/reserves/KyberFprReserveV2.sol
Anyhowclick: image

This will be an additional 36-37k gas for FPRs, and 7k-84k for APRs. It's especially bad for t2e rate queries, and token decimals aren't 18.

contracts/sol6/reserves/KyberFprReserveV2.sol
Anyhowclick: Ah yeah, that will suffice. None that I can think of for now... @ayobuenavista @ssreelekhya any other feedback from ur interactions with the teams you've onboarded regarding reserves?

contracts/sol6/reserves/KyberFprReserveV2.sol
Anyhowclick: image

This will be an additional 36-37k gas for FPRs, and 7k-90k for APRs. It's especially bad for t2e rate queries, and token decimals aren't 18.

ilanDoron: how many steps does each source quantity have for fpr in this table?

package.json
manhlx3006: Updated

contracts/sol6/reserves/KyberFprReserveV2.sol
manhlx3006: I've added rate validation in trade func, some gas data from tests (average 15 transactions):

With rate validation in trade function:
1. Eth + Tokens in reserve: 114,000
2. Eth in Reserve, Token in wallet: 122,000
3. Weth: 147,000 (increase around 33k due to transfer + wrap/unwrap)

Without rate validation in trade function:
1. Eth + Tokens in reserve: 75,300
2. Eth in Reserve, Token in wallet: 82,000
3. Weth: 106,000 (increase around 31k due to transfer + wrap/unwrap)

Summary:
1. Using weth costs around 30k-35k more gas for each trade due to transfer + wrap/unwrap
2. Rate validation again in trade function costs 40k more, (on average, depends on number of steps, one step could cost around 2k-3k more according to data from conversion rate contract)

ilanDoron: OK

Weth costs pretty much.
rate validation around 40k matches in large results from Desmond above.
but seems one step is less then 2-3k

just need to see how much it would be with max steps.
which is 10 qty 10 imbalance for conversionRate
and 16 steps for enhanced conversionRate

contracts/sol6/reserves/IConversionRates.sol
```diff
@@ -0,0 +1,22 @@
+pragma solidity 0.6.6;
+
```
ilanDoron, line 2: please use existing file
IKyberSanity.sol

contracts/sol6/reserves/IWeth.sol
```diff
@@ -0,0 +1,9 @@
+pragma solidity 0.6.6;
+
+import "../IERC20.sol";
```
ilanDoron, line 3: wondering.
does this file belong in reserves folder.
can we include it maybe from open zeppelin?

test/sol6/reserves/kyberFprReserveV2.js
```diff
@@ -0,0 +1,2991 @@
...
+const reserveSetup = require("../../reserveSetup.js");
+const BN = web3.utils.BN;
+
+//global variables
+//////////////////
+const {precisionUnits, ethDecimals, ethAddress, zeroAddress, zeroBN, MAX_ALLOWANCE} = require("../../helper.js");
+const {expectEvent, expectRevert} = require('@openzeppelin/test-helpers');
+
+//balances
+let expectedReserveBalanceWei = new BN(0);
```
ilanDoron, line 18: should we have these as global parameters?
thinking we should aim for each test to be able to run separate.
in this case no sense of following reserve balance between tests.

contracts/sol6/reserves/KyberFprReserveV2.sol
manhlx3006: With/without rate validation:
- No sanity rate:
    - Different for 0 step: 26.7k
    - Gas consumption per step: 2k
    - Min: 26.7 (0 step), Max: 26.7 + 2 * 20 = 66.7k (20 steps)
- Has sanity rate:
    - Different for 0 steps: 31.8k
        - 26.7k for calling conversion rate 0 step
        - 5.1k extra for sanity rate call
    - Gas consumption per step: 2k
    - Min: 31.8k (0 step), Max: 31.8 + 2 * 20 = 71.8k (20 steps)

ilanDoron: OK
we continue with the rate validation
later we should find a way to save gas for this validation

test/sol6/reserves/kyberFprReserveV2.js
ilanDoron: yes
not relevant.

test/sol6/reserves/kyberFprReserveV2.js
```diff
@@ -0,0 +1,3060 @@
...
+let maxGasPrice = new BN(150).mul(new BN(10).pow(new BN(9))); // 150 * 10^9
+
+//tokens data
+////////////
+let numTokens = 5;
+let tokens = [];
+let tokenDecimals = [];
+let tokenAdd = [];
+
+// all price steps in bps (basic price steps).
```
ilanDoron, line 46: I think this comment doesn't belong here.

its good to have it in other places.
like dev portal or conversion rate test file.

contracts/sol6/reserves/KyberFprReserveV2.sol
```diff
@@ -0,0 +1,357 @@
...
+            // add to imbalance
+            conversionRatesContract.recordImbalance(
+                destToken,
+                int256(destAmount),
+                0,
+                block.number
+            );
+            // if reserve is using weth, convert eth to weth and transfer weth to its' tokenWallet
+            if (srcTokenWallet != address(this)) {
+                weth.deposit{value: msg.value}();
```
ilanDoron, line 325: so actually the assumption here is that if its weth its external wallet.
and if its an external for Eth its using WETH

I mean
no option to use weth and have wallet address to be this.

its good. 
just not sure its totally clear.

I think we can add another @dev note in setTokenWallet

contracts/sol6/reserves/KyberFprReserveV2.sol
```diff
@@ -0,0 +1,357 @@
...
+        approvedWithdrawAddresses[keccak256(abi.encodePacked(address(token), addr))] = approve;
+        setDecimals(token);
+        emit WithdrawAddressApproved(token, addr, approve);
+    }
+
+    /// @dev allow set tokenWallet[token] back to 0x0 address
+    function setTokenWallet(IERC20 token, address wallet) external onlyAdmin {
+        tokenWallet[address(token)] = wallet;
+        setDecimals(token);
+        emit NewTokenWallet(token, wallet);
```
ilanDoron, line 123: should we add a special event for eth?
also wondering should it warn if reserve has eth balance?

test/reserveSetup.js
```diff
@@ -0,0 +1,195 @@
...
+let validRateDurationInBlocks = 1000;
+let maxGasPrice = new BN(150).mul(new BN(10).pow(new BN(9))); // 150 * 10^9
+
+// imbalance data
+let minimalRecordResolution = new BN(2);
+let maxPerBlockImbalance = 40000;
+let maxTotalImbalance = maxPerBlockImbalance * 12;
+
+
+//imbalance buy steps
```
ilanDoron, line 22: lets update this to a more traditional setup
so buy imbalance x values are only positive
and sell imbalance x values are negative.

also nice to add the option to pass in steps input

manhlx3006: Ok, for steps input can just modify this initial array data

ilanDoron: OK

contracts/sol6/reserves/KyberFprReserveV2.sol
manhlx3006: There is no reason of using weth in this reserve address right? we can not approve other wallets to get weth from reserve

manhlx3006: I will add a dev note in setTokenWallet

contracts/sol6/reserves/KyberFprReserveV2.sol
```diff
@@ -0,0 +1,357 @@
...
+            }
+        }
+
+        emit WithdrawFunds(token, amount, destination);
+    }
+
+    function setContracts(
+        address _kyberNetwork,
+        IConversionRates _conversionRates,
+        IWeth _weth,
```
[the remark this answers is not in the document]
manhlx3006: It is unchanged now, but not sure later, just keep the flexibility

contracts/sol6/reserves/KyberFprReserveV2.sol
```diff
@@ -0,0 +1,357 @@
...
+        IKyberSanity _sanityRates
+    ) external onlyAdmin {
```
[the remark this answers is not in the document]
manhlx3006: yeah ok we should have separate set address, so we don't have to find the others if only update 1

contracts/sol6/reserves/KyberFprReserveV2.sol
```diff
@@ -0,0 +1,357 @@
...
+        emit DepositToken(ETH_TOKEN_ADDRESS, msg.value);
+    }
+
+    function trade(
+        IERC20 srcToken,
+        uint256 srcAmount,
+        IERC20 destToken,
+        address payable destAddress,
+        uint256 conversionRate,
+        bool validate
```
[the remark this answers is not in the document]
manhlx3006: ok

test/sol6/reserves/kyberFprReserveV2.js
manhlx3006: ok will remove from reserve test file

test/sol6/reserves/kyberFprReserveV2.js
manhlx3006: all tests can run separately, these params are global so we don't have to define for each test suite or pass them in validate function

contracts/sol6/reserves/KyberFprReserveV2.sol
ilanDoron: yes
no reason.
just to clarify

contracts/sol6/reserves/KyberFprReserveV2.sol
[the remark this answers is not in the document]
ilanDoron: yes
would be better.

test/sol6/reserves/kyberFprReserveV2.js
ilanDoron: Ahh
I see

contracts/sol6/reserves/KyberFprReserveV2.sol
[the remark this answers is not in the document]
ilanDoron: OK

test/sol6/reserves/kyberFprReserveV2.js
```diff
@@ -0,0 +1,2972 @@
...
+            await token.approve(reserve.address, 0, {from: walletForToken});
+            await token.transfer(walletForToken, new BN(10));
+            Helper.assertEqual(0, await reserve.getBalance(tokenAdd[tokenInd]));
+
+            // approve less than balance
+            await token.approve(reserve.address, new BN(2), {from: walletForToken});
+            Helper.assertEqual(new BN(2), await reserve.getBalance(tokenAdd[tokenInd]));
+
+            // approve more than balance
+            await token.approve(reserve.address, 0, {from: walletForToken});
```
ducquangkstn, line 2309: redundant

contracts/sol6/reserves/KyberFprReserveV2.sol
```diff
@@ -0,0 +1,358 @@
...
+        address payable destAddress,
+        uint256 conversionRate
+    ) internal {
+        require(conversionRate > 0, "rate is 0");
+
+        uint256 rate;
+
+        if (srcToken == ETH_TOKEN_ADDRESS) {
+            require(msg.value == srcAmount, "wrong msg value");
+            rate = conversionRatesContract.getRate(destToken, block.number, true, srcAmount);
```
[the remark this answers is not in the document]
ilanDoron: Good idea.
its depends on trust model
@tranvictor can u respond?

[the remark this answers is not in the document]
tranvictor: My short answer is No. It is not the interest of the reserves, they always want to be safer and they don't have to spend gas on the validation. Why would they set it to false?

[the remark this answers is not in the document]
ilanDoron: they could set it to false for supporting lower kyber gas prices.
its true they don't pay
but they do have the interest to keep kyber gas prices more competitive


and it does turn to "semi trusted" relations
they can always turn it on if they suspect something.
And we at Kyber can't control this. or don't always know the actual setup.

contracts/sol6/reserves/KyberFprReserveV2.sol
```diff
@@ -0,0 +1,358 @@
...
+        } else {
+            // add to imbalance
+            conversionRatesContract.recordImbalance(
+                srcToken,
+                -1 * int256(srcAmount),
+                0,
+                block.number
+            );
+            // collect src token from sender
+            srcToken.safeTransferFrom(msg.sender, srcTokenWallet, srcAmount);
```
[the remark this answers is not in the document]
ilanDoron: we can still fail after line 229 due to:
sanity rate
zero dest amount 
rate parameter too high.

if we transfer tokens on 229 and fail. user spends more gas.
But I do agree the above failures are with almost zero likelihood.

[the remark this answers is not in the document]
manhlx3006: Here is also clear right? You collect srcToken, then transfer destToken to recipient

contracts/sol6/reserves/KyberFprReserveV2.sol
```diff
@@ -0,0 +1,357 @@
...
+        );
+    }
+
+    function getConversionRate(
+        IERC20 src,
+        IERC20 dest,
+        uint256 srcQty,
+        uint256 blockNumber
+    ) external override view returns (uint256) {
+        if (!configData.tradeEnabled) return 0;
```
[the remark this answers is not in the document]
manhlx3006: Checked with some local tests, it is 1 sload for configData

contracts/sol6/reserves/KyberFprReserveV2.sol
manhlx3006: I added the dev comment, reserve manager that uses this reserve should have knowledge of how it works

ilanDoron: OK
good enough.

contracts/sol6/reserves/KyberFprReserveV2.sol
```diff
@@ -0,0 +1,389 @@
...
+
+        if (doRateValidation) {
+            uint256 rate = conversionRatesContract.getRate(
+                isBuy ? destToken : srcToken,
+                block.number,
+                isBuy,
+                srcAmount
+            );
+            // re-validate conversion rate
+            require(rate >= conversionRate, "rate is lower than conversion rate");
```
ilanDoron, line 333: maybe write

reserve rate lower then network requested rate

contracts/sol6/reserves/KyberFprReserveV2.sol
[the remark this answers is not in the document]
ilanDoron: Great
thanks

contracts/sol6/reserves/KyberFprReserveV2.sol
```diff
@@ -0,0 +1,389 @@
...
+    /// @param destToken Destination token
+    /// @param destAddress Destination address to send tokens to
+    /// @param doRateValidation do rate validation or not
+    function doTrade(
+        IERC20 srcToken,
+        uint256 srcAmount,
+        IERC20 destToken,
+        address payable destAddress,
+        uint256 conversionRate,
+        bool doRateValidation
```
ilanDoron, line 314: compiler warning



contracts/sol6/reserves/KyberFprReserveV2.sol:314:9: Warning: This declaration shadows an existing declaration.
        bool doRateValidation
        ^-------------------^
contracts/sol6/reserves/KyberFprReserveV2.sol:257:5: The shadowed declaration is here:
    function doRateValidation() external view returns (bool) {
    ^ (Relevant source part starts here and spans across multiple lines).

contracts/sol6/reserves/KyberFprReserveV2.sol
```diff
@@ -0,0 +1,389 @@
...
+        require(tx.gasprice <= uint256(configData.maxGasPriceWei), "gas price too high");
+
+        doTrade(
+            srcToken,
+            srcAmount,
+            destToken,
+            destAddress,
+            conversionRate,
+            configData.doRateValidation
+        );
```
ilanDoron, line 101: tested with remix
See this file.
https://gist.github.com/ilanDoron/5ffd42ff4424d17e2eb41e3df1344e16

seems trade2 is best for gas.

ilanDoron (commented):
Accidentally two other comments are not in this code review.
pressed the wrong button.
so please see 3 comments.
note checks are failing.

contracts/sol6/reserves/KyberFprReserveV2.sol
manhlx3006: hm I see, the second was even higher in the js test I wrote, but it seems to be lower when testing with real deployed contract. Let me change to second method

test/sol6/reserves/kyberFprReserveV2.js
```diff
@@ -0,0 +1,3422 @@
...
+                user1,
+                false
+              );
+              numberTxsPerStep[tx.numberSteps]++;
+              gasUsedPerStep[tx.numberSteps].iadd(new BN(tx.gasUsed));
+            }
+
+            for(let i = 0; i <= 10; i++) {
+              if (numberTxsPerStep[i] > 0) {
+                console.log(`         Average gas used for buy with ${i} steps: ${Math.floor(1.0 * gasUsedPerStep[i]/numberTxsPerStep[i])}`)
```
ilanDoron, line 3146: seems here it should be 
used for sell (not buy)

test/reserveSetup.js
```diff
@@ -0,0 +1,409 @@
...
+    let compactSellArr = [];
+
+    let qtyBuyStepX = [0, 5000, 10000, 16000, 28000, 32000, 45000];
+    let qtyBuyStepY = [ 0, 10, 20, 32, 44, 56, 101];
+    let qtySellStepX = [0, 5000, 11000, 18000, 30000, 36000, 48000];
+    let qtySellStepY = [ 0, 12, 22, 33, 45, 58, 110];
+    let imbalanceBuyStepX = [0, 5000, 15000, 21000, 28000, 33000, 45000];
+    let imbalanceBuyStepY = [0, -110, -160, -250, -440, -1000, -1600];
+    let imbalanceSellStepX = [-48000, -39000, -25000, -19000, -14000, -10000, 0];
+    let imbalanceSellStepY = [-1500, -1200, -440, -320, -200, -75, 0];
```
ilanDoron, line 39: I see in some tests its with 10 steps
Are steps configured in a different place in the js file?

test/sol6/reserves/kyberFprReserveV2.js
manhlx3006: ok

test/reserveSetup.js
manhlx3006: for gas consumption v1 tests, the number of steps reported are total of quantity and imbalance steps. For v2, it is only imbalanace step

contracts/sol6/reserves/KyberFprReserveV2.sol
manhlx3006: change to `validateRate`

contracts/sol6/reserves/KyberFprReserveV2.sol
manhlx3006: ok, update the revert msg

test/sol6/reserves/kyberFprReserveV2.js
```diff
@@ -0,0 +1,3422 @@
...
+              }
+            }
+          });
+
+          it("Test few sells with steps", async() => {
+            //init conversion rate
+            await setupConversionRateV2sContractWithStepsV1();
+            await generalSetupReserveContract(isUsingWallet[t], isUsingWeth[t]);
+
```
ilanDoron, line 3362: better add this setup for all sell runs

await reserveInst.setDoRateValidation(isDoRateValidation[r], {from: admin});


I think its not set well.
and also important if someone running a separate test.

test/reserveSetup.js
ilanDoron: Ahhh got it.
OK

test/sol6/reserves/kyberFprReserveV2.js
ilanDoron: notice its also missing in other sell steps loop.

contracts/sol6/reserves/KyberFprReserveV2.sol
ilanDoron: maybe the 3rd parameter in the struct made some difference

test/sol6/reserves/kyberFprReserveV2.js
```diff
@@ -0,0 +1,3422 @@
...
+		withdrawAddress = accounts[6];
+		alerter = accounts[8];
+		walletForToken = accounts[9];
+
+		currentBlock = await Helper.getCurrentBlock();
+		weth = await WethToken.new("WrapETH", "WETH", 18);
+
+		// create tokens
+		for (let i = 0; i < numTokens; ++i) {
+			tokenDecimals[i] = new BN(18);
```
ilanDoron, line 85: somehow here its linted with 4 spaces tab.
and later with 2.

can run prettier on full file.

test/reserveSetup.js
```diff
@@ -0,0 +1,409 @@
...
+        bps: (imbalanceBuyStepY[imbalanceBuyStepY.length - 1]),
+        steps: imbalanceBuyStepY.length
+    }
+};
+
+module.exports.getExtraBpsForImbalanceSellQuantityV1 = function(qty, imbalanceSellStepX, imbalanceSellStepY) {
+    for (let i = 0; i < imbalanceSellStepX.length; i++) {
+        if (qty <= imbalanceSellStepX[i]) {
+            return {
+                bps: imbalanceSellStepY[i],
```
ilanDoron, line 400: please lint with prettier.

test/sol6/reserves/kyberFprReserveV2.js
manhlx3006: Updated, it was missing in sell loops

Anyhowclick (approved):

